PR #1, opened 2021-05-15
fix kine dependency

balchua:
The `kine` dependency is pointing to a local directory that does not exist on others.


balchua:
@ktsakalozos PTAL

PR #13, opened 2021-08-23
Downgrade kine to what we know it works

ktsakalozos:
Downgrade kine to v0.4.1 from canonical
Clean-up go dependencies
Removed server_test as it was pulling an outdated grpc dependency.


joedborg (approved):
LGTM

PR #43, opened 2023-03-22
TLS dial ClientSessionCache size configuration

neoaggelos:
### Summary

Make the ClientSessionCache configurable in the used TLS dial configuration. Setting a positive size should improve performance with TLS session resumption, but comes at the cost of being incompatible with previous go-dqlite versions.

### Testing

- (nodeA) Install MicroK8s on 1.26/stable
- (nodeB) Install MicroK8s on latest/edge with this build and --tls-client-session-cache-size=256
- Join nodeB to nodeA. The join process stalls, from the logs we observe that dqlite cannot communicate
- Remove the `--tls-client-session-cache-size` from nodeB, using the default size 0
- Join completes without issues

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -32,13 +32,14 @@ import (
 )
 
 var opts = options.Options{
-	StorageDir:      "/var/tmp/k8s-dqlite",
-	ListenEp:        "tcp://127.0.0.1:12379",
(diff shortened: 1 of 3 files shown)


ktsakalozos (approved):
LGTM +1

PR #53, opened 2023-08-24
[WIP] Make go-migrator a part of k8s-dqlite

neoaggelos:
### Summary

Build `go-migrator` as a sub-command of `k8s-dqlite`. Results in a reduced snap size and less binaries to build/bundle.


neoaggelos:
The implementation is based on https://github.com/canonical/go-migrator/blob/master/main.go. Note that we drop the `direct` method, since it is not used anywhere in MicroK8s

PR #66, opened 2023-09-21
Added full disk watcher and startup condition

berkayoz:
This would be a starting point to handle full disk cases. The conditions can checked by a watcher which will try to gracefully handover leadership/voter rights and shutdown. The syscall for `unix.Statfs` is really fast, so this check can be handled on queries that would put data to the disk such as `CREATE`,`UPDATE` and `DELETE`.  If having a watcher seems to be the better idea, this work can be finalized by cleaning up some of the ugly code.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -32,6 +32,7 @@ var (
 
 		validateAvailableStorageInterval time.Duration
 		validateAvailableStorageMinBytes uint64
+		lowAvailableStorageAction        string
 	}
 
 	rootCmd = &cobra.Command{
@@ -70,6 +71,7 @@ var (
 				rootCmdOpts.minTLSVersion,
 				rootCmdOpts.validateAvailableStorageInterval,
 				rootCmdOpts.validateAvailableStorageMinBytes,
+				rootCmdOpts.lowAvailableStorageAction,
 			)
 			if err != nil {
 				logrus.WithError(err).Fatal("Failed to create server")
@@ -125,6 +127,7 @@ func init() {
 	rootCmd.Flags().StringVar(&rootCmdOpts.minTLSVersion, "min-tls-version", "tls12", "Minimum TLS version for dqlite endpoint (tls10|tls11|tls12|tls13). Default is tls12")
 	rootCmd.Flags().BoolVar(&rootCmdOpts.metrics, "metrics", true, "enable metrics endpoint")
 	rootCmd.Flags().StringVar(&rootCmdOpts.metricsAddress, "metrics-listen", "127.0.0.1:9042", "listen address for metrics endpoint")
-	rootCmd.Flags().DurationVar(&rootCmdOpts.validateAvailableStorageInterval, "validate-storage-available-size-interval", 0, "Interval to check if the disk is running low on space. Set to 0 to disable the periodic disk size check")
-	rootCmd.Flags().Uint64Var(&rootCmdOpts.validateAvailableStorageMinBytes, "validate-storage-available-size-min-bytes", 10*1024*1024, "Minimum required available disk size (in bytes) to continue operation. If available disk space gets below this threshold, the server will refuse to start")
+	rootCmd.Flags().DurationVar(&rootCmdOpts.validateAvailableStorageInterval, "validate-storage-available-size-interval", 5*time.Second, "Interval to check if the disk is running low on space. Set to 0 to disable the periodic disk size check")
+	rootCmd.Flags().Uint64Var(&rootCmdOpts.validateAvailableStorageMinBytes, "validate-storage-available-size-min-bytes", 10*1024*1024, "Minimum required available disk size (in bytes) to continue operation. If available disk space gets below this threshold, then the --low-available-storage-action is performed")
+	rootCmd.Flags().StringVar(&rootCmdOpts.lowAvailableStorageAction, "low-available-storage-action", "none", "Action to perform in case the available storage is low. One of (none|handover|terminate). none means no action is performed. handover means the dqlite node will handover its leadership role, if any. terminate means this dqlite node will shutdown")
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -29,6 +29,8 @@ var (
 		minTLSVersion          string
 		metrics                bool
 		metricsAddress         string
+		storageWatchPeriod     uint16
+		storageDiskThreshold   uint64
 	}
 
 	rootCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -41,6 +41,18 @@ var (
 			if rootCmdOpts.debug {
 				logrus.SetLevel(logrus.TraceLevel)
 			}
+			var stat unix.Statfs_t
+
+			unix.Statfs(rootCmdOpts.dir, &stat)
+
+			avail := stat.Bavail * uint64(stat.Bsize)
+
+			var disk_threshold uint64 = 10000000
```
ktsakalozos, line 50: Could we have this configurable, as an argument. We should have also the query interval as an argument.

pkg/server/server.go
```diff
@@ -249,6 +255,36 @@ func (s *Server) Start(ctx context.Context) error {
...
+				unix.Statfs(s.storageDir, &stat)
+				avail := stat.Bavail * uint64(stat.Bsize)
+				// If available space is less than 10MB reject to start.
+				if avail < disk_threshold {
+					// Create a separate context with 30 seconds to cleanup
+					stopCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+					defer cancel()
+
+					logrus.Debug("Handing over dqlite leadership")
+					if err := s.app.Handover(stopCtx); err != nil {
```
ktsakalozos, line 272: What happens if you are not a leader?

pkg/server/server.go
```diff
@@ -37,6 +38,10 @@ var expectedFilesDuringInitialization = map[string]struct{}{
 	"tuning.yaml":    {},
 }
 
+var storageWatchTicker = time.NewTicker(5 * time.Second)
+var storageWatchQuit = make(chan struct{})
+var disk_threshold uint64 = 10000000
```
bschimke95, line 43: Let's reuse the value from above and not define it twice.

pkg/server/server.go
```diff
@@ -16,13 +16,14 @@ import (
...
+	"golang.org/x/sys/unix"
 )
 
 // Server is the main k8s-dqlite server.
 type Server struct {
 	// app is the dqlite application driving the server.
-	app *app.App
-
+	app        *app.App
+	storageDir string
```
bschimke95, line 26: Please add a comment

pkg/server/server.go
```diff
@@ -249,6 +255,36 @@ func (s *Server) Start(ctx context.Context) error {
 	}
 	logrus.WithFields(logrus.Fields{"id": s.app.ID(), "address": s.app.Address()}).Print("Started dqlite")
 
+	go func() {
+		for {
+			select {
+			case <-storageWatchTicker.C:
+				var stat unix.Statfs_t
+				unix.Statfs(s.storageDir, &stat)
```
bschimke95, line 263: You could move this logic to a util as this logic is also used in the other file.
e.g. 
```
if util.diskIsFull() {
...

File: pkg/server/server.go [hunk 6]
```diff
@@ -249,6 +255,36 @@ func (s *Server) Start(ctx context.Context) error {
...
+				avail := stat.Bavail * uint64(stat.Bsize)
+				// If available space is less than 10MB reject to start.
```
bschimke95, line 265: This comment will quickly run out of sync - IMHO better not add the concrete threshold here.

bschimke95 (commented):
Nice work! Only a few comments.
Also, can we add a small test to cover that logic?

pkg/server/server.go
berkayoz: Go-dqlite handles this in the background, only voter and leadership rights are handed over. Please see [app.go#L311-L316](https://github.com/canonical/go-dqlite/blob/9e965d50ce079bea63e8c76b18c07f3ca91fc2e4/app/app.go#L311-L316)

pkg/server/server.go
```diff
@@ -237,18 +244,55 @@ func New(dir string, listen string, enableTLS bool, diskMode bool, clientSession
...
+					logrus.WithError(err).Errorf("failed to close dqlite app")
+				}
+				logrus.WithField("dir", s.swatcherConfig.storageDir).Fatalf("Disk is critically low, gracefully shutdown...")
+			}
+		case <-s.swatcherConfig.storageWatchQuit:
+			s.swatcherConfig.storageWatchTicker.Stop()
+			return
+		}
+	}
+}
```
neoaggelos, line 285: This should instead accept the `context.Context` passed in start, and return in `case <-ctx.Done()`

pkg/server/server.go
```diff
@@ -237,18 +244,55 @@ func New(dir string, listen string, enableTLS bool, diskMode bool, clientSession
...
+
 // Start the dqlite node and the kine machinery.
 func (s *Server) Start(ctx context.Context) error {
 	if err := s.app.Ready(ctx); err != nil {
 		return fmt.Errorf("failed to start dqlite app: %w", err)
 	}
 	logrus.WithFields(logrus.Fields{"id": s.app.ID(), "address": s.app.Address()}).Print("Started dqlite")
 
+	go s.WatchStorage()
```
neoaggelos, line 294: Let's pass the context here

cmd/root.go
```diff
@@ -42,6 +44,14 @@ var (
 				logrus.SetLevel(logrus.TraceLevel)
 			}
 
+			isFull, err := server.IsStorageFull(rootCmdOpts.dir, rootCmdOpts.storageDiskThreshold)
+			// If available space is less than threshold reject startup.
+			if err != nil {
+				logrus.WithError(err).Fatal("failed to check storage capacity")
+			} else if isFull {
+				logrus.WithField("dir", rootCmdOpts.dir).Fatal("Disk is critically low, rejecting startup")
```
neoaggelos, line 52:
```suggestion
				logrus.WithField("dir", rootCmdOpts.dir).Fatal("Disk is critically low, refusing to start")
```

pkg/server/server.go
```diff
@@ -237,18 +244,55 @@ func New(dir string, listen string, enableTLS bool, diskMode bool, clientSession
...
+			storageDir:         dir,
+			storageWatchTicker: time.NewTicker(time.Duration(storageWatchPeriod) * time.Second),
+			storageWatchQuit:   make(chan struct{}),
+			diskThreshold:      storageDiskThreshold,
+		},
 		kineConfig: kineConfig,
 	}, nil
 }
 
+func (s *Server) WatchStorage() {
```
neoaggelos, line 258: This could be `watchStorage` instead, unless we need to export this?

pkg/server/server.go
```diff
@@ -237,18 +244,55 @@ func New(dir string, listen string, enableTLS bool, diskMode bool, clientSession
...
+				defer cancel()
+				logrus.Debug("Handing over dqlite leadership")
+				if err := s.app.Handover(stopCtx); err != nil {
+					logrus.WithError(err).Errorf("Failed to handover dqlite")
+				}
+				logrus.Debug("Closing dqlite application")
+				if err := s.app.Close(); err != nil {
+					logrus.WithError(err).Errorf("failed to close dqlite app")
+				}
+				logrus.WithField("dir", s.swatcherConfig.storageDir).Fatalf("Disk is critically low, gracefully shutdown...")
```
neoaggelos, line 278: This should instead be handled by the context cancellation, we should not rewrite parts of `Shutdown()`.

I would recommend adding a `stopCh chan(struct{})` to the `Server` struct, then:

```
func (s *Server) Done() <-chan struct{} {
  return s.stopCh
}
```

And update the root command to wait for a cancellation like this:

```go
select {
case <-ch:  // received a signal
case <-server.Done(): // the server decided to terminate
}
```

neoaggelos (commented):
Looks OK overall, leaving some comments to keep things tidy

PR #68, opened 2023-10-05
Add admission control

bschimke95:
This PR adds a way to fail requests early based on admission policies which reduces the load on the database.
The following policies are currently available:
    
* `allow-all`: Disables admission control
* `limit`: Denies queries if maximum amount of Txn is reached.

Closes MK-1302

### DIFF
--- a/pkg/kine/drivers/generic/admission.go
+++ b/pkg/kine/drivers/generic/admission.go
@@ -1,0 +1,48 @@
+package generic
+
+import (
+	"context"
+	"fmt"
+
+	"golang.org/x/sync/semaphore"
+)
+
+// AdmissionControlPolicy interface defines the admission policy contract.
+type AdmissionControlPolicy interface {
+	// Admit checks whether the query should be admitted.
+	// If the query is not admitted, a non-nil error is returned with the reason why the query was denied.
+	// If the query is admitted, then the error will be nil and a callback function is returned to the caller.
+	// The caller must execute it after finishing the query
+	Admit(context.Context) (callOnFinish func(), err error)
+}
+
+// allowAllPolicy always admits queries.
+type allowAllPolicy struct{}
+
+// Admit always admits requests for AllowAllPolicy.
+func (p *allowAllPolicy) Admit(context.Context) (func(), error) {
+	return func() {}, nil
+}
+
+// limitPolicy denies queries when the maximum threshold is reached.
+type limitPolicy struct {
+	MaxRunningTxn int64
+	semaphore     *semaphore.Weighted
+}
+
+func newLimitPolicy(maxRunningTxn int64) *limitPolicy {
+	return &limitPolicy{
+		MaxRunningTxn: maxRunningTxn,
+		semaphore:     semaphore.NewWeighted(maxRunningTxn),
+	}
+}
+
+func (p *limitPolicy) admit(ctx context.Context) (func(), error) {
+	ok := p.semaphore.TryAcquire(1)
+	if !ok {
+		return func() {}, fmt.Errorf("current Txns reached limit (%d)", p.MaxRunningTxn)
+	}
+	return func() {
+		p.semaphore.Release(1)
+	}, nil
+}
--- a/pkg/kine/drivers/generic/admission_test.go
+++ b/pkg/kine/drivers/generic/admission_test.go
@@ -1,0 +1,43 @@
+package generic
+
+import (
+	"context"
+	"testing"
+
+	"github.com/onsi/gomega"
+)
+
+func TestAllowAllPolicy_Admit(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+	policy := &allowAllPolicy{}
+
+	callOnFinish, err := policy.Admit(context.Background())
+
+	g.Expect(err).To(gomega.BeNil())
+	g.Expect(callOnFinish).ToNot(gomega.BeNil())
+}
+
+func TestLimitPolicy_Admit(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+	policy := newLimitPolicy(2)
+
+	// First two should succeed
+	done1, err := policy.admit(context.Background())
+	g.Expect(err).ToNot(gomega.HaveOccurred())
+	g.Expect(done1).ToNot(gomega.BeNil())
+	done2, err := policy.admit(context.Background())
+	g.Expect(err).ToNot(gomega.HaveOccurred())
+	g.Expect(done2).ToNot(gomega.BeNil())
+
+	// Third should be denied
+	done3, err := policy.admit(context.Background())
+	g.Expect(err).To(gomega.HaveOccurred())
+	// should still return a valid function as callers otherwise might segfault if they not check for nil.
+	g.Expect(done3).ToNot(gomega.BeNil())
+
+	// Complete a call - now the next query should be admitted again.
+	done1()
+	_, err = policy.admit(context.Background())
+	g.Expect(err).ToNot(gomega.HaveOccurred())
+	g.Expect(done3).ToNot(gomega.BeNil())
+}
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -123,6 +123,8 @@ type Generic struct {
 	TranslateErr                  TranslateErr
 	ErrCode                       ErrCode
 
+	ac AdmissionControlPolicy
+
 	// CompactInterval is interval between database compactions performed by kine.
 	CompactInterval time.Duration
 	// PollInterval is the event poll interval used by kine.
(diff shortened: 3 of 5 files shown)


pkg/kine/drivers/generic/generic.go
```diff
@@ -315,8 +320,16 @@ func (d *Generic) query(ctx context.Context, txName, sql string, args ...interfa
 		if err != nil {
 			err = fmt.Errorf("query (try: %d): %w", i, err)
 		}
+		d.ac.Metrics.NumCurrentTxn.Add(-1)
 		recordOpResult(txName, err, start)
 	}()
+
+	d.ac.Metrics.NumCurrentTxn.Add(1)
```
bschimke95, line 327: I'm not really happy about how this is implemented. I'd rather have the admission control only once but could not really figure out a unique entry point for all requests where this could be done (without refactoring large portions of the code.

Happy to hear opinions/alternative suggestions.

pkg/kine/drivers/generic/generic.go
```diff
@@ -240,6 +242,9 @@ func Open(ctx context.Context, driverName, dataSourceName string, paramCharacter
 
 		FillSQL: q(`INSERT INTO kine(id, name, created, deleted, create_revision, prev_revision, lease, value, old_value)
 			VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)`, paramCharacter, numbered),
+
+		// TODO(MK-1397): Make policy and config configurable via CLI.
+		ac: NewAdmissionController(&LimitPolicy{MaxRunningTxn: 5000}),
```
bschimke95, line 247: Maybe we should use 
```
ac: NewAdmissionController(&AllowAllPolicy{}),
```
here to disable this feature until https://warthogs.atlassian.net/browse/MK-1397 is implemented? I have not yet checked the typical amount of Txn when running microk8s (this is part of https://warthogs.atlassian.net/browse/MK-1398).

pkg/kine/drivers/generic/generic.go
```diff
@@ -311,12 +317,21 @@ func getPrefixRange(prefix string) (start, end string) {
 func (d *Generic) query(ctx context.Context, txName, sql string, args ...interface{}) (rows *sql.Rows, err error) {
 	i := uint(0)
 	start := time.Now()
+
+	done, err := d.ac.Admit(ctx)
+	defer done()
+	if err != nil {
+		err = fmt.Errorf("denied: %w", err)
+		return
+	}
```
neoaggelos, line 326: I don't like this `defer done()`, can we instead call it specifically when we want to, right after the SQL transaction is complete?

neoaggelos (changes requested):
Looking better, left some more comments

pkg/kine/drivers/generic/generic.go
```diff
@@ -471,8 +511,17 @@ func (d *Generic) Count(ctx context.Context, prefix string) (int64, int64, error
 
 func (d *Generic) CurrentRevision(ctx context.Context) (int64, error) {
 	var id int64
+	var err error
+
+	done, err := d.ac.Admit(ctx)
+	if err != nil {
+		err = fmt.Errorf("denied: %w", err)
+		return 0, err
```
neoaggelos, line 519: also in other places

```suggestion
		return 0, fmt.Errorf("denied: %w", err)
```

pkg/kine/drivers/generic/generic.go
```diff
@@ -325,6 +339,7 @@ func (d *Generic) query(ctx context.Context, txName, sql string, args ...interfa
 			logrus.Tracef("QUERY (try: %d) %v : %s", i, args, strippedSQL)
 		}
 		rows, err = d.DB.QueryContext(ctx, sql, args...)
+		done()
```
neoaggelos, line 342: Do we release on every loop here? This does not seem right

pkg/kine/drivers/generic/generic.go
```diff
@@ -337,8 +352,17 @@ func (d *Generic) query(ctx context.Context, txName, sql string, args ...interfa
 
 func (d *Generic) queryPrepared(ctx context.Context, txName, sql string, prepared *sql.Stmt, args ...interface{}) (result *sql.Rows, err error) {
 	logrus.Tracef("QUERY %v : %s", args, Stripped(sql))
+
+	done, err := d.ac.Admit(ctx)
+	if err != nil {
+		err = fmt.Errorf("denied: %w", err)
+		return
```
neoaggelos, line 359: also in other places 
```suggestion
		return nil, fmt.Errorf("denied: %w", err)
```

pkg/kine/drivers/generic/admission_test.go
```diff
@@ -0,0 +1,47 @@
...
+		policy.semaphore.Acquire(context.Background(), 2)
+		defer policy.semaphore.Release(2)
+		policy.metrics.NumCurrentTxn.Store(2)
+
+		callOnFinish, err := policy.Admit(context.Background())
+
+		g.Expect(err).To(gomega.HaveOccurred())
+		g.Expect(callOnFinish).ToNot(gomega.BeNil())
+		g.Expect(policy.metrics.NumCurrentTxn.Load()).To(gomega.Equal(int64(2)))
+	})
```
neoaggelos, line 46: This seems to be very closely tied to the implementation itself. How about a test where we test the functionality instead?

e.g: on limit 2, first two succeed, third one fails, one completes then a new one can succeed?

neoaggelos (commented):
Getting there

pkg/kine/drivers/generic/generic.go
```diff
@@ -329,16 +342,26 @@ func (d *Generic) query(ctx context.Context, txName, sql string, args ...interfa
 			time.Sleep(jitter.Deviation(nil, 0.3)(2 * time.Millisecond))
 			continue
 		}
+		done()
```
neoaggelos, line 345: I am not sure I get this. Why do we call `done()` on every iteration?

bschimke95: This part of the code tries to query the DB. If it fails it will retry in a loop 500 times. So, this `done` call will only be reached once the DB query finishes without error (otherwise the loop will continue due to the `continue` call on line 343.

To cover both cases, successful DB query and failed query after 500 tries, we need to call the `done` method once in the loop and once on line 349.

neoaggelos: Ah, I see now, thanks

pkg/kine/drivers/generic/admission.go
```diff
@@ -0,0 +1,60 @@
...
+// LimitPolicy denies queries when the maximum threshold is reached.
+type LimitPolicy struct {
+	metrics       LimitPolicyMetrics
+	MaxRunningTxn int64
+	semaphore     *semaphore.Weighted
+}
+
+func NewLimitPolicy(maxRunningTxn int64) *LimitPolicy {
+	return &LimitPolicy{
+		MaxRunningTxn: maxRunningTxn,
```
neoaggelos, line 44: any reason why this should be public?

pkg/kine/drivers/generic/admission.go
```diff
@@ -0,0 +1,60 @@
...
+// Admit always admits requests for AllowAllPolicy.
+func (p *AllowAllPolicy) Admit(context.Context) (func(), error) {
+	return func() {}, nil
+}
+
+// AdmissionMetrics stores admission-related metrics.
+// TODO(MK-1398): Integrate those metrics with Prometheus.
+type LimitPolicyMetrics struct {
+	// NumCurrentTxn stores the number of currently running transactions.
+	NumCurrentTxn atomic.Int64
```
neoaggelos, line 32: any reason why this should be public? let's maybe fully remove the unused metrics atm, and work on adding a prometheus counter in a next PR

test/mixed_workload_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"testing"
+	"time"
+
+	. "github.com/onsi/gomega"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+// TestMixed puts heavy load on kine and expects that some requests are denied
+// by the admission control.
+func TestMixed(t *testing.T) {
```
neoaggelos, line 17: Let's then maybe call this `TestAdmissionControl`?

neoaggelos (commented):
Looks good overall, some final comments

pkg/kine/drivers/generic/admission.go
bschimke95: Good point - removed

test/mixed_workload_test.go
bschimke95: Done

pkg/kine/drivers/generic/admission.go
bschimke95: Nope - I was used to have file = package and forgot that this is also contained in the `generic` package. Thanks

berkayoz (commented):
Looks good overall, great work!

Leaving it to @neoaggelos for approval.

test/mixed_workload_test.go
```diff
@@ -0,0 +1,108 @@
+package test
```
neoaggelos, line 1: Perhaps also rename file to something `test/admission_test.go`?

neoaggelos (approved):
LGTM, one final comment

test/mixed_workload_test.go
bschimke95: Done

PR #72, opened 2023-10-12
Add `--admission-control-only-write-queries` flag

bschimke95:
If this flag is enabled, the admission control will only perform checks on write queries. Read queries will be ignored.

### DIFF
--- a/pkg/kine/drivers/generic/admission.go
+++ b/pkg/kine/drivers/generic/admission.go
@@ -25,12 +25,12 @@ const (
 )
 
 // Queries that perform write operations on the database
-var writeQueries = map[string]struct{}{
-	"update_compact_sql":        {},
-	"delete_sql":                {},
-	"fill_sql":                  {},
-	"insert_last_insert_id_sql": {},
-	"insert_sql":                {},
+var writeQueries = map[string]bool{
+	"update_compact_sql":        true,
+	"delete_sql":                true,
+	"fill_sql":                  true,
+	"insert_last_insert_id_sql": true,
+	"insert_sql":                true,
 }
 
 // allowAllPolicy always admits queries.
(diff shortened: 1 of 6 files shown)


pkg/kine/drivers/generic/admission.go
```diff
@@ -18,16 +18,27 @@ type AdmissionControlPolicy interface {
...
-	statusAccepted string = "accepted"
-	statusDenied   string = "denied"
+	// operation was evaluated by the admission control and accepted
+	resultAccepted string = "accepted"
+	// operation was evaluated by the admission control and denied by policy
+	resultDenied string = "denied"
 )
 
+// Queries that perform write operations on the database
+var writeQueries = map[string]struct{}{
```
neoaggelos, line 28: Having a second look at it, this could also be a `map[string]bool`. That way we can use the fact that looking up a missing key will return a zero value.

pkg/kine/drivers/generic/admission.go
```diff
@@ -38,33 +49,46 @@ func (p *allowAllPolicy) Admit(ctx context.Context, txName string) (func(), erro
...
+	_, isWriteQuery := writeQueries[txName]
+	return !p.onlyWriteQueries || isWriteQuery
 }
 
 func (p *limitPolicy) Admit(ctx context.Context, txName string) (func(), error) {
-	ok := p.semaphore.TryAcquire(1)
-	if !ok {
-		recordOpAdmissionControl(txName, statusDenied)
-		return func() {}, fmt.Errorf("number of concurrent database operations reached limit (%d)", p.maxConcurrentTxn)
+	if p.shouldCheck(txName) {
```
neoaggelos, line 70: With this, we will evaluate `shouldCheck` twice for each query. Once here, and once in the callback. This is not ideal, since this is supposed to be on the hot path.

Further, I am not in favor of a function `shouldCheck`, as this does not give the caller any info about what it actually checks.

Given that we only need to do this once, we could just do the bool expression, with a comment explaining it (if needed)

```go
// we need to lock the semaphore if ...
needSem := !onlyWriteQueries || isWriteQuery[tx_name]
```

neoaggelos (changes requested):
Almost there

neoaggelos (changes requested):
Almost there

pkg/kine/drivers/generic/admission.go
bschimke95: Done

pkg/kine/drivers/generic/admission.go
bschimke95: Done

bschimke95 (commented):
Thanks for the comments, applied the changes

neoaggelos (approved):
LGTM, nice!

PR #74, opened 2023-10-16
Added write batching on Insert(LogStructured Append) calls

berkayoz:
Added write batching on Insert(LogStructured Append) calls

This will need some more iterations and cleaning up to make it a feature and definitely needs tests to verify functionality.
However, through manual testing, the Append calls seem to work as expected.

The batching adds more latency to inserts, with the average max being `batching-interval`. 
Sweetspot for `batching-interval` and `batching-max-queries` is critical for optimum outcome. Having these values not optimal will have negative performance impacts.

### DIFF
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -90,6 +90,11 @@ type ErrRetry func(error) bool
 type TranslateErr func(error) error
 type ErrCode func(error) string
 
+type BatchedInsert struct {
+	args  []interface{}
+	retCh chan int64
+}
+
 type Generic struct {
 	sync.Mutex
 
@@ -122,13 +127,18 @@ type Generic struct {
 	Retry                         ErrRetry
 	TranslateErr                  TranslateErr
 	ErrCode                       ErrCode
+	flushCh                       chan struct{}
+	batchingQueue                 []BatchedInsert
 
 	AdmissionControlPolicy AdmissionControlPolicy
 
 	// CompactInterval is interval between database compactions performed by kine.
 	CompactInterval time.Duration
 	// PollInterval is the event poll interval used by kine.
-	PollInterval time.Duration
+	PollInterval       time.Duration
+	BatchingInterval   time.Duration
+	BatchingMaxQueries int
+	BatchingEnabled    bool
 }
 
 func configureConnectionPooling(db *sql.DB) {
@@ -381,6 +391,104 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
 	return r
 }
 
+func (d *Generic) InitializeWriteBatching(ctx context.Context) {
+	if d.BatchingEnabled && d.BatchingInterval > 0 && d.BatchingMaxQueries > 0 {
+		d.flushCh = make(chan struct{})
+		d.batchingQueue = make([]BatchedInsert, 0, d.BatchingMaxQueries)
+
+		go d.watchBatchQueue(ctx)
+	}
+}
+
+func (d *Generic) processBatchedInserts(ctx context.Context, queue []BatchedInsert) ([]int64, error) {
+	valueStrings := make([]string, 0, len(queue))
+	// len(queue) * argument_count per insert
+	valueArgs := make([]interface{}, 0, len(queue)*8)
+
+	for _, insertItem := range queue {
+		valueStrings = append(valueStrings, "(?, ?, ?, ?, ?, ?, ?, ?)")
+		valueArgs = append(valueArgs, insertItem.args...)
+	}
+
+	// Maybe move this into the Generic struct like other queries
+	// Although since we are dynamically generating this we'd only store the template string there
+	stmt := fmt.Sprintf("INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value) VALUES%s", strings.Join(valueStrings, ","))
+
+	prepared, err := d.DB.Prepare(stmt)
+	if err != nil {
+		logrus.WithError(err).Error("Process batch queue error")
+		return nil, err
+	}
+
+	row, err := d.executePrepared(ctx, "batch_insert_last_insert_id_sql", stmt, prepared, valueArgs...)
+
+	if err != nil {
+		logrus.WithError(err).Error("Process batch queue error")
+		return nil, err
+	}
+
+	lastId, err := row.LastInsertId()
+
+	if err != nil {
+		logrus.WithError(err).Error("Process batch queue error")
+		return nil, err
+	}
+
+	var ids []int64
+
+	// Ideally instead of counting back from LastInsertId, sqlite supports RETURNING
+	// so something like "INSERT INTO .... RETURNING" id would be better.
+	// However db.Prepare complains about wrong syntax when RETURNING is used
+	for i := len(queue); i > 0; i-- {
+		ids = append(ids, lastId-int64(i)+1)
+	}
+
+	return ids, nil
+}
+
+func (d *Generic) processBatchQueue(ctx context.Context) {
+	if len(d.batchingQueue) > 0 {
+		dcq := make([]BatchedInsert, len(d.batchingQueue))
+		copy(dcq, d.batchingQueue)
+
+		ids, err := d.processBatchedInserts(ctx, dcq)
+
+		if err != nil {
+			logrus.WithError(err).Error("Process batch queue error")
+			return
+		}
+
+		for i, insertItem := range dcq {
+			insertItem.retCh <- ids[i]
+		}
+
+		d.batchingQueue = make([]BatchedInsert, 0, d.BatchingMaxQueries)
+	}
+}
+
+func (d *Generic) watchBatchQueue(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			d.processBatchQueue(ctx)
+			return
+		case <-d.flushCh:
+			d.processBatchQueue(ctx)
+		case <-time.After(d.BatchingInterval):
+			d.processBatchQueue(ctx)
+		}
+	}
+}
+
+func (d *Generic) batchQueryRowPrepared(ctx context.Context, txName, sqli string, prepared *sql.Stmt, args ...interface{}) (id int64) {
+	ret := make(chan int64)
+	d.batchingQueue = append(d.batchingQueue, BatchedInsert{retCh: ret, args: args})
+	if len(d.batchingQueue) == d.BatchingMaxQueries {
+		d.flushCh <- struct{}{}
+	}
+	return <-ret
+}
+
 func (d *Generic) executePrepared(ctx context.Context, txName, sql string, prepared *sql.Stmt, args ...interface{}) (result sql.Result, err error) {
 	i := uint(0)
 	start := time.Now()
@@ -563,6 +671,12 @@ func (d *Generic) Insert(ctx context.Context, key string, create, delete bool, c
 	}
 	if delete {
 		dVal = 1
+	}
+
+	if d.BatchingEnabled {
+		// Down below should return errors encountered while the batched operations are processed
+		id = d.batchQueryRowPrepared(ctx, "insert_sql", d.InsertSQL, d.insertSQLPrepared, key, cVal, dVal, createRevision, previousRevision, ttl, value, prevValue)
+		return id, nil
 	}
 
 	if d.LastInsertID {
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -450,11 +450,18 @@ func (d *Generic) batchQueueWatcher(ctx context.Context) {
 }
 
 func (d *Generic) processBatchedInserts(ctx context.Context, queue []BatchedInsert) ([]int64, error) {
+	_, rev, err := d.GetCompactRevision(ctx)
+	if err != nil {
+		logrus.WithError(err).Errorf("Get current rev error")
+		return nil, err
+	}
+
 	valueStrings := make([]string, 0, len(queue))
 	// len(queue) * argument_count per insert
 	valueArgs := make([]interface{}, 0, len(queue)*8)
 
-	for _, insertItem := range queue {
+	for i, insertItem := range queue {
+		insertItem.args[4] = rev + int64(i+1)
 		valueStrings = append(valueStrings, "(?, ?, ?, ?, ?, ?, ?, ?)")
 		valueArgs = append(valueArgs, insertItem.args...)
 	}
@@ -723,7 +730,6 @@ func (d *Generic) Insert(ctx context.Context, key string, create, delete bool, c
 	}
 
 	if d.BatchingEnabled {
-		// Down below should return errors encountered while the batched operations are processed
 		id, err := d.batchQueryRowPrepared(ctx, "insert_sql", d.InsertSQL, d.insertSQLPrepared, key, cVal, dVal, createRevision, previousRevision, ttl, value, prevValue)
 		return id, err
 	}
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -314,7 +314,7 @@ func (d *Generic) query(ctx context.Context, txName, sql string, args ...interfa
 	i := uint(0)
 	start := time.Now()
 
-	done, err := d.AdmissionControlPolicy.Admit(ctx)
+	done, err := d.AdmissionControlPolicy.Admit(ctx, txName)
 	if err != nil {
 		return nil, fmt.Errorf("denied: %w", err)
 	}
@@ -349,7 +349,7 @@ func (d *Generic) query(ctx context.Context, txName, sql string, args ...interfa
 func (d *Generic) queryPrepared(ctx context.Context, txName, sql string, prepared *sql.Stmt, args ...interface{}) (result *sql.Rows, err error) {
 	logrus.Tracef("QUERY %v : %s", args, Stripped(sql))
 
-	done, err := d.AdmissionControlPolicy.Admit(ctx)
+	done, err := d.AdmissionControlPolicy.Admit(ctx, txName)
 	if err != nil {
 		return nil, fmt.Errorf("denied: %w", err)
 	}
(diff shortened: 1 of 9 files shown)


pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +391,104 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		if err != nil {
+			logrus.WithError(err).Error("Process batch queue error")
+			return
+		}
+
+		for i, insertItem := range dcq {
+			insertItem.retCh <- ids[i]
+		}
+
+		d.batchingQueue = make([]BatchedInsert, 0, d.BatchingMaxQueries)
```
bschimke95, line 465: This introduces a race issue if there are write calls coming in while processing the queue.
The problem is that the whole queue is cleared and not just the ones that were processed. (see branch `MK-1417/batching-test` `batching_test.go` to reproduce this)

I fixed this with 
```
d.batchingQueue = d.batchingQueue[len(dcq):]
```

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +393,119 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+
+		d.removeFromBatchQueueCh <- len(dcq)
+	}
+}
+
+func (d *Generic) watchBatchQueue(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			d.processBatchQueue(ctx)
```
neoaggelos, line 493: This will attempt to process the queue but will use a cancelled context, and so it will most likely fail.

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +393,119 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+	stmt := fmt.Sprintf("INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value) VALUES%s", strings.Join(valueStrings, ","))
+
+	prepared, err := d.DB.Prepare(stmt)
+	if err != nil {
+		logrus.WithError(err).Error("Process batch queue error")
+		return nil, err
+	}
+
+	row, err := d.executePrepared(ctx, "batch_insert_last_insert_id_sql", stmt, prepared, valueArgs...)
+
```
neoaggelos, line 445: I am not sure if we should prepare a statement that is immediately dropped afterwards.

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +393,119 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		ids = append(ids, lastId-int64(i)+1)
+	}
+
+	return ids, nil
+}
+
+func (d *Generic) processBatchQueue(ctx context.Context) {
+	if len(d.batchingQueue) > 0 {
+		dcq := make([]BatchedInsert, len(d.batchingQueue))
+		copy(dcq, d.batchingQueue)
```
neoaggelos, line 473: This happens in parallel, would also need locking.

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +393,119 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		select {
+		case <-ctx.Done():
+			return
+		case item := <-d.addToBatchQueueCh:
+			d.batchingQueue = append(d.batchingQueue, item)
+			if len(d.batchingQueue) == d.BatchingMaxQueries {
+				d.flushCh <- struct{}{}
+			}
+		case count := <-d.removeFromBatchQueueCh:
+			d.batchingQueue = d.batchingQueue[count:]
```
neoaggelos, line 419: These most likely need a mutex to protect

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +393,119 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		case <-time.After(d.BatchingInterval):
+			d.processBatchQueue(ctx)
+		}
+	}
+}
+
+func (d *Generic) batchQueryRowPrepared(ctx context.Context, txName, sqli string, prepared *sql.Stmt, args ...interface{}) (id int64) {
+	ret := make(chan int64)
+	d.addToBatchQueueCh <- BatchedInsert{retCh: ret, args: args}
+	return <-ret
```
neoaggelos, line 506: we must `close(ret)` this after we are done, or we are leaking channels

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +393,119 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
 	return r
 }
 
+func (d *Generic) InitializeWriteBatching(ctx context.Context) {
+	if d.BatchingEnabled && d.BatchingInterval > 0 && d.BatchingMaxQueries > 0 {
```
neoaggelos, line 397: We are adding a new feature, I would recommending validating the config and failing early if it is not proper (e.g. if batchingMaxQueries or batchingInterval are negative, let's shout the error instead of silently disabling this.

I would also like a logging message like this:

```go
logrus.WithFields(logrus.Fields{"interval": interval, "queue_size": maxQueries}).Print("Enable batching of insert queries")
```

or

```
logrus.Print("Disable batching of insert queries")
```

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +393,119 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		case <-d.flushCh:
+			d.processBatchQueue(ctx)
+		case <-time.After(d.BatchingInterval):
+			d.processBatchQueue(ctx)
+		}
+	}
+}
+
+func (d *Generic) batchQueryRowPrepared(ctx context.Context, txName, sqli string, prepared *sql.Stmt, args ...interface{}) (id int64) {
+	ret := make(chan int64)
```
neoaggelos, line 504: So that we don't block when writing to the channel

```suggestion
	ret := make(chan int64, 1)
```

neoaggelos (commented):
I very much like how this looks, but this needs work, leaving some initial comments for discussion.

pkg/kine/drivers/generic/generic.go
neoaggelos: Maybe also something similar to https://github.com/canonical/k8s-dqlite/blob/0ce314e0424d3e2cf744c64cd4e6d1b3d215cfb9/pkg/server/server.go#L253

bschimke95: We also need an upper limit for the max-queue parameter. Otherwise we can run into 
```
time="2023-10-20T11:34:40+02:00" level=error msg="Process batch queue error" error="too many SQL variables"
```
when creating the query.
See also: https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber

pkg/kine/drivers/generic/generic.go
bschimke95: Just to learn something: Why do we need to protect the queue here? We only change the queue through this watcher and therefore the changes should not happen concurrently, right?

pkg/kine/drivers/generic/generic.go
```diff
@@ -394,15 +396,32 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+	if !d.BatchingEnabled {
+		return
 	}
+
+	if d.BatchingInterval < 10*time.Millisecond {
+		logrus.WithFields(logrus.Fields{"interval": d.BatchingInterval, "queue_size": d.BatchingMaxQueries}).Print("Interval is less than 10ms, disabling batching of insert queries")
+		return
+	}
+
+	if d.BatchingMaxQueries < 5 {
```
bschimke95, line 409: We also need an upper limit - up until ~120 works fine for me. After that I get 
```
time="2023-10-20T16:14:45+02:00" level=error msg="Process batch queue error" error="too many SQL variables"
``` errors

File: pkg/kine/drivers/generic/generic.go [hunk 10]
```diff
@@ -394,15 +396,32 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
-		d.removeFromBatchQueueCh = make(chan int)
-		d.batchingQueue = make([]BatchedInsert, 0, d.BatchingMaxQueries)
 
-		go d.watchBatchQueue(ctx)
-		go d.batchQueueWatcher(ctx)
+	if !d.BatchingEnabled {
+		return
 	}
+
+	if d.BatchingInterval < 10*time.Millisecond {
```
bschimke95, line 404: It would make sense to give an upper limit here as well to:
Otherwise inexperienced users might set high intervals which might block the calls for too long.

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +395,142 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		logrus.WithFields(logrus.Fields{"interval": d.BatchingInterval, "queue_size": d.BatchingMaxQueries}).Print("Interval is less than 10ms, disabling batching of insert queries")
+		return
+	}
+
+	if d.BatchingMaxQueries < 5 {
+		logrus.WithFields(logrus.Fields{"interval": d.BatchingInterval, "queue_size": d.BatchingMaxQueries}).Print("Queue size is less than 5, disabling batching of insert queries")
+		return
+	}
```
neoaggelos, line 412: I would recommended returning an error and failing instead

pkg/kine/drivers/generic/generic.go
```diff
@@ -411,12 +430,13 @@ func (d *Generic) batchQueueWatcher(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case item := <-d.addToBatchQueueCh:
-			d.batchingQueue = append(d.batchingQueue, item)
-			if len(d.batchingQueue) == d.BatchingMaxQueries {
+			d.batchingCond.L.Lock()
+			if len(d.batchingQueue) >= d.BatchingMaxQueries {
```
bschimke95, line 434: This shouldn't be necessary if synchronization is working properly, right?

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +395,142 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		case <-ctx.Done():
+			return
+		case item := <-d.addToBatchQueueCh:
+			d.batchingCond.L.Lock()
+			if len(d.batchingQueue) >= d.BatchingMaxQueries {
+				d.flushCh <- struct{}{}
+				d.batchingCond.Wait()
+			}
+			d.batchingQueue = append(d.batchingQueue, item)
+			d.batchingCond.L.Unlock()
```
neoaggelos, line 439: What does this offer us over a simple `batchingQueueMu sync.Mutex` ?

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +395,142 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+		select {
+		case <-ctx.Done():
+			return
+		case item := <-d.addToBatchQueueCh:
+			d.batchingCond.L.Lock()
+			if len(d.batchingQueue) >= d.BatchingMaxQueries {
+				d.flushCh <- struct{}{}
+				d.batchingCond.Wait()
+			}
+			d.batchingQueue = append(d.batchingQueue, item)
```
ktsakalozos, line 438: The unlucky query that will trigger the flushing will have to wait two flush cycles, one to empty the queue and a second to flush the queue it was placed to. What if we first append the query and then see if we need to flush?

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +395,142 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+	if len(d.batchingQueue) > 0 {
+		d.batchingCond.L.Lock()
+		dcq := make([]BatchedInsert, len(d.batchingQueue))
+		copy(dcq, d.batchingQueue)
+		d.batchingCond.L.Unlock()
+
+		ids, err := d.processBatchedInserts(ctx, dcq)
+		if err != nil {
+			logrus.WithError(err).Error("Process batch queue error")
+			return
```
ktsakalozos, line 500: If there is an error shouldn't we let all the batched queries waiting for a result that they have failed? Why don't we we have the retCh in the insetItem have an error too?

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +395,142 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+	// Although since we are dynamically generating this we'd only store the template string there
+	stmt := fmt.Sprintf("INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value) VALUES%s", strings.Join(valueStrings, ","))
+
+	prepared, err := d.DB.Prepare(stmt)
+	if err != nil {
+		logrus.WithError(err).Error("Process batch queue error")
+		return nil, err
+	}
+
+	row, err := d.executePrepared(ctx, "batch_insert_last_insert_id_sql", stmt, prepared, valueArgs...)
```
ktsakalozos, line 464: If we get two insert requests with the same key we fail with:
```
Οκτ 21 20:27:17 aurora microk8s.daemon-k8s-dqlite[234689]: time="2023-10-21T20:27:17+03:00" level=error msg="Process batch queue error" error="exec (try: 0): UNIQUE constraint failed: kine.prev_revision, kine.name"
```

ktsakalozos: The prev_revision should also be unique, probably

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +395,142 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+
+	if err != nil {
+		logrus.WithError(err).Error("Process batch queue error")
+		return nil, err
```
ktsakalozos, line 468: When a request fails we should remove it from the queue.

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +395,142 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+	// so something like "INSERT INTO .... RETURNING" id would be better.
+	// However db.Prepare complains about wrong syntax when RETURNING is used
+	for i := len(queue); i > 0; i-- {
+		ids = append(ids, lastId-int64(i)+1)
+	}
+
+	return ids, nil
+}
+
+func (d *Generic) processBatchQueue(ctx context.Context) {
```
ktsakalozos, line 490: When executing the processBatchQueue we lock the queue, create a copy of the items we want to processBatchedInserts then we unlock the queue and we do the processing. At the end of the processing we lock the queue and we remove the section that we processed. This processBatchQueue is called (from watchBatchQueue) either if we flush or if the BatchingInterval expires. It should be possible to have multiple calls processing the same items from the queue. While the first call is in processBatchedInserts a second can enter the processBatchQueue and take the same queue items to process.

pkg/kine/drivers/generic/generic.go
berkayoz: `condition.Wait()` let's us pause the queue adding goroutine(unlocks the mutex and pauses) when the queue is full and wait's for a signal to wake up which is raised when the queue is consumed at the processing goroutine. Not pausing the queue adding goroutine leads to queries overflowing the max queue size.

pkg/kine/drivers/generic/generic.go
berkayoz: This seems tricky, initial plan would be maybe try to adjust the `prev_revision` and increment by `i` in `range queue`. But multiple queue processing cycles seems to come with the same revision. The reason being the revision is fetched from the table before they are processed. Maybe keeping the revision in memory instead of fetching from the db and incrementing accordingly might help although I'm not sure how it would play out on the HA scenario.

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +400,151 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+	const maxInterval = 500 * time.Millisecond
+	const minQueueSize = 5
+	const maxQueueSize = 100
+
+	if !d.BatchingEnabled {
+		return nil
+	}
+
+	if d.BatchingInterval < minInterval || d.BatchingInterval > maxInterval {
+		return errors.New("Interval should be between 5ms to 500ms, disabling batching of insert queries")
```
bschimke95, line 414: Please use the `minInterval` and `maxInterval` variables

pkg/kine/drivers/generic/generic.go
```diff
@@ -381,6 +400,151 @@ func (d *Generic) queryRowPrepared(ctx context.Context, txName, sql string, prep
...
+	}
+
+	if d.BatchingMaxQueries < minQueueSize || d.BatchingMaxQueries > maxQueueSize {
+		return errors.New("Queue size should be between 5 to 100, disabling batching of insert queries")
```
bschimke95, line 418: Please use the variables defines above.

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -216,6 +226,24 @@ func parseOpts(dsn string) (opts, error) {
...
+		case "batching-interval":
+			d, err := time.ParseDuration(vs[0])
+			if err != nil {
+				return opts{}, fmt.Errorf("failed to parse batching-interval duration value %q: %w", vs[0], err)
+			}
+			result.batchingInterval = d
+		case "batching-max-queries":
+			d, err := strconv.ParseInt(vs[0], 10, 64)
+			if err != nil {
+				return opts{}, fmt.Errorf("failed to parse max-concurrent-txn value %q: %w", vs[0], err)
```
bschimke95, line 238: Replace copy-pasted value

bschimke95 (commented):
In the standup you said something about a write/update tests that you implemented to validate the `prev_revision` changes. Could you add this test to the PR please?

PR #92, opened 2024-02-28
1.29 conformance fixes

neoaggelos:
## Summary

Update k8s-dqlite to pass the Kubernetes 1.29 conformance tests

## Changes

- Compaction only ignores the last 100 revisions (from 1000).
- List responses return keys in alphabetical order.
- List requests with paging now return the correct `Count` of results that remain
- List requests with paging properly handle the start key.
- Properly handle list responses where the keys revisions are not in alphabetical order.

See the individual commits for details. Each commit is accompanied by a test case that doesn't work without the introduced changes.


ktsakalozos (approved):
LGTM +1

PR #107, opened 2024-06-25
Bench time workload

marco6:
This PR aims at measuring time only for the workload of the benchmarks and ignore the setup work. This way we should get more accurate results to compare.


marco6:
Depends on #106

marco6:
Marking as ready since #106 was merged.

bschimke95 (approved):
LGTM

PR #109, opened 2024-06-27
Fix migration test

marco6:
The initial schema for the migration test was missing a column and the test was failing.


bschimke95 (approved):
LGTM, thanks for the fix

marco6:

PR #110, opened 2024-06-27
Improve retry logic

marco6:
During high contention SQLite will sometime return `SQLITE_BUSY`. In those cases, the driver should retry the query for some arbitrary number of times, before giving up, to make the query more reliable.

This PR adds both the logic for sqlite driver and the generic retry logic for `GetCompactRevision` which was missing it.


bschimke95 (approved):
LGTM

marco6:

PR #111, opened 2024-06-27
Make DoCompact operation interruptible

marco6:
This PR aims at making the `DoCompact` operation interruptible from the outside, so that it is easier to test.


bschimke95 (approved):
LGTM

marco6:

PR #114, opened 2024-07-02
Add synchronization primitive for server shutdown

marco6:
This PR adds a way for the Server to wait for other services (like the Backend, the event polling, the cleanup for ttl, the watchers, etc...) to end during shutdown.

It is useful mainly during tests/benchmarks so that we are sure that once the test is over, no service is still running.


bschimke95 (approved):
LGTM

PR #115, opened 2024-07-03
Remove UpdateSameKeyLinearity

marco6:
This test is too imprecise to be useful and making it precise is unpractical. Right now if the second batch takes *twice* the time (leading to an exponential growth) the test still pass.

It is also the test that takes the longest amount of time in the whole pipeline.


bschimke95 (approved):
LGTM

PR #116, opened 2024-07-04
Simplify integration tests

marco6:
This PR aims at reducing the boilerplate for tests and benchmarks. In that direction it:
 - aligns the names of the utility function (now consistently called `{create,update,delete}Key`)
 - tries to use them everywhere possible
 - aligns the usage of gomega for the admission test
 - removes unused helper functions

From a functionality level, it also removes same-nesting inter-test dependencies which might break the tests if the run filters them (very present in the list tests). This has also a nice effect of reducing a bit indentation.

PR #117, opened 2024-07-04
Remove build flag

marco6:
This PR aims at simplifying the run of tests and make sure each test runs only once during the CI pipeline. For this reason, it removes the capability of building this project without dqlite.

The reason for this change is that it is useful to run test with a single command and generate a single report (for coverage, for benchmarks, etc...).

It is still possible (while not very useful) to run only dqlite/sqlite tests by filtering with the `-run` flag like:

```
go test -v ./... -run dqlite
```


marco6:
Depends on #116

marco6:
Marking as ready since #116 was merged.

bschimke95 (approved):
LGTM

PR #120, opened 2024-07-08
Stable benchmarks

marco6:
This PR makes each benchmark run at least 7 times (which in general is the minimum amount of repetitions to compute `p` with `benchstat` tool).

Given that this will increase again bench time, I removed the building of dqlite and used the PPA instead, so that some of the time spent there can be saved.


marco6:
Draft as depends on #117

marco6:
Marking as ready since #117 was merged.

bschimke95 (approved):
LGTM

PR #121, opened 2024-07-09
Suppress logrus output during tests

marco6:
Logrus output during tests is flooding the output and making it difficult to properly run `benchstat` on the output. This PR redirects the output to `io.Discard` for that and sets the log level to `Fatal` so that the log footprint is minimal.


bschimke95 (approved):
LGTM

PR #122, opened 2024-07-09
Add continuous benchmarking

marco6:
This PR gather results from the benchmarking job and compares them with the base branch, so that we can monitor how a change affects performance.

It mostly uses standard tooling, so the output is still a bit rough, but I think it's for the best as it's the same format one would read while running benchmarks on his own machine.


marco6:
Superseeded by #123

PR #125, opened 2024-07-10
Instrument Otel

louiseschmidtgen:
### Description
This PR introduces otel tracing logs.

## Traces
We collect traces on (get/count/create/update/delete/list/update/compact/dbsize), this includes various child traces where relevant.
When you run `/hack/jaeger.sh` and run `k8s-dqlite` with `--otel=true`, the traces are collected in a local jaeger instance at `http://localhost:16686`.

Screenshot from 2024-07-10 10-14-56

Screenshot from 2024-07-10 10-15-25

## Metrics
The metric counters (get/count/create/update/delete/list/update) are written to stdout every 30 seconds.
(Limitation: Jaeger can't accept these metrics)
```
{
      "Name": "k8s-dqlite.get",
      "Description": "Number of get requests",
      "Unit": "",
      "Data": {
        "DataPoints": [
          {
            "Attributes": [],
            "StartTime": "2024-07-10T09:52:13.907275271+02:00",
            "Time": "2024-07-10T10:24:13.907264437+02:00",
            "Value": 1685
          }
        ],
        "Temporality": "CumulativeTemporality",
        "IsMonotonic": true
```
@bschimke95 

## Existing metrics
Side-note: Our kine layer already tracks some metrics here: `http://127.0.0.1:9042/metrics`:
See an example output [here](https://pastebin.canonical.com/p/htssK8WSN2/)


pkg/kine/server/create.go
```diff
@@ -28,13 +29,25 @@ func (l *LimitedServer) create(ctx context.Context, put *etcdserverpb.PutRequest
 		return nil, unsupported("prevKv")
 	}
 
+	ctx, span := tracer.Start(ctx, "backend.create")
+	defer span.End()
+
+	span.SetAttributes(
+
```
bschimke95, line 36: nit: remove empty line

pkg/kine/server/get.go
```diff
@@ -5,18 +5,33 @@ import (
...
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.opentelemetry.io/otel/attribute"
 )
 
 func (l *LimitedServer) get(ctx context.Context, r *etcdserverpb.RangeRequest) (*RangeResponse, error) {
 	if r.Limit != 0 && len(r.RangeEnd) != 0 {
 		return nil, fmt.Errorf("invalid combination of rangeEnd and limit, limit should be 0 got %d", r.Limit)
 	}
 
+	ctx, span := tracer.Start(ctx, "backend.get")
```
bschimke95, line 16: looks like you use both -  `backend.x` and `limited.x`. Any reason for this or should we settle on one version?

pkg/kine/server/tracing.go
```diff
@@ -0,0 +1,126 @@
...
+	// handleErr calls shutdown for cleanup and makes sure that all errors are returned.
+	handleErr := func(inErr error) {
+		err = errors.Join(inErr, shutdown(ctx))
+	}
+
+	// Set up resource.
+	res, err := resource.New(ctx,
+		resource.WithAttributes(
+			// The service name used to display traces in backends
+			semconv.ServiceNameKey.String("k8s-dqlite"),
```
bschimke95, line 46: you can use the const `name` defined in `otel` here.

pkg/kine/server/update.go
```diff
@@ -31,10 +32,28 @@ func (l *LimitedServer) update(ctx context.Context, rev int64, key string, value
...
+		ctx, span := tracer.Start(ctx, "backend.update")
+		defer span.End()
+		span.SetAttributes(
+			attribute.String("key", key),
+			attribute.Int64("revision", rev),
+			attribute.Int64("lease", lease),
+		)
 		rev, kv, ok, err = l.backend.Update(ctx, key, value, rev, lease)
+		span.SetAttributes(attribute.Bool("ok", ok))
+		updateCnt.Add(ctx, 1)
```
bschimke95, line 56: You sometimes increase the counter before performing the action and somtimes afterwards. I see that this would not make a difference here *because there is no way to return early) but let's be consistent.

bschimke95 (commented):
Some initial comments, looks good so far

pkg/kine/server/get.go
louiseschmidtgen: Yes, so limited is the limitedserver.x call which then makes a call to backend.x, or backend.y based on the logic.

louiseschmidtgen: Screenshot from 2024-07-10 12-31-45

eaudetcobello:
Nice work

dimaqq:
Fly-by comment: I've like to see some docs about where the parent span comes from, env? header field? etc. Additionally, given the distributed nature of dqlite, how spans in dlite peers may be linked to each other.

louiseschmidtgen:
> Fly-by comment: I've like to see some docs about where the parent span comes from, env? header field? etc. Additionally, given the distributed nature of dqlite, how spans in dlite peers may be linked to each other.

@dimaqq Great question! 
To answer your first question, we retrieve information about the parent span from the context header fields. See otel's [spanContext](https://opentelemetry.io/docs/specs/otel/trace/api/#spancontext), they follow the [W3C TraceContext Specification](https://www.w3.org/TR/trace-context/#sampled-flag).
In summary, a `traceparent header` has the values `span-id` `trace-id` and `trace-flags (sampled flag)`. Additionally, the `tracestate header` is used which allows for multiple tracing systems in a trace.

louiseschmidtgen:
Here are some screenshots from traces I found interesting/have questions about:
Does revision 0 mean a deleted entry? Why is there a count 9 on it?
Screenshot from 2024-07-11 20-49-41

Backend updates run very frequently in an idle cluster. Interestingly, lease is always zero for these. A create query returns a lease > 0.
Screenshot from 2024-07-11 15-27-07

Idle cluster runs constant update queries:
Screenshot from 2024-07-10 10-22-12


This shows the revision range that we run two consecutive compactions on:
Screenshot from 2024-07-11 21-25-09

bschimke95 (commented):
sorry, forgot to press the submit button.

Only a minor comment afterwards we are good to go.

pkg/kine/logstructured/sqllog/otel.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
+)
+
+const name = "sqllog"
+
+var (
+	tracer     = otel.Tracer(name)
```
marco6, line 13: The initialization in this file is mixed (some are in the `init`, some are in the global). Can we uniform that?

pkg/kine/logstructured/sqllog/otel.go
```diff
@@ -0,0 +1,20 @@
...
+	meter      = otel.Meter(name)
+	compactCnt metric.Int64Counter
+)
+
+func init() {
+	compactCnt, _ = meter.Int64Counter(fmt.Sprintf("%s.compact", name), metric.WithDescription("Number of compact requests"))
```
marco6, line 19: What happens if there is an error here? I think that `compactCnt` will be `nil` and so the code will panic on its first usage. I think we should panic here, if an error occurs, instead of having an error down the line.

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -124,17 +124,25 @@ func (s *SQLLog) DoCompact(ctx context.Context) error {
 }
 
 func (s *SQLLog) compactor(ctx context.Context, nextEnd int64) (int64, error) {
+	_, span := tracer.Start(ctx, "sqllog.compactor")
```
marco6, line 127: I think we should use the returned context, so that if we create a span in `CurrentRevision` and `GetCompactRevision` (or any other method) it will be grouped under `sqllog.compactor`.

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -124,17 +124,25 @@ func (s *SQLLog) DoCompact(ctx context.Context) error {
...
+	defer span.End()
+	span.AddEvent(fmt.Sprintf("nextEnd: %d", nextEnd))
```
marco6, line 129: Could we use attributes here instead of formatting a string? I see that in `limitedserver` we are doing that

pkg/kine/server/get.go
```diff
@@ -5,18 +5,33 @@ import (
...
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.String("key", string(r.Key)),
+		attribute.String("rangeEnd", string(r.RangeEnd)),
+		attribute.Int64("limit", r.Limit),
+		attribute.Int64("revision", r.Revision),
+	)
 	if r.Limit != 0 && len(r.RangeEnd) != 0 {
+		span.RecordError(fmt.Errorf("invalid combination of rangeEnd and limit, limit should be 0 got %d", r.Limit))
```
marco6, line 22: Can we use defer from that and not duplicate this error?

pkg/kine/server/create.go
```diff
@@ -28,13 +36,18 @@ func (l *LimitedServer) create(ctx context.Context, put *etcdserverpb.PutRequest
 		return nil, unsupported("prevKv")
 	}
 
+	createCnt.Add(ctx, 1)
+
 	rev, err := l.backend.Create(ctx, string(put.Key), put.Value, put.Lease)
+	span.SetAttributes(attribute.Int64("revision", rev))
 	if err == ErrKeyExists {
+		span.RecordError(err)
```
marco6, line 44: This shouldn't be recorded as it is not a misbehavior.

pkg/kine/server/create.go
```diff
@@ -28,13 +36,18 @@ func (l *LimitedServer) create(ctx context.Context, put *etcdserverpb.PutRequest
...
 		return &etcdserverpb.TxnResponse{
 			Header:    txnHeader(rev),
 			Succeeded: false,
 		}, nil
 	} else if err != nil {
+		span.RecordError(err)
```
marco6, line 50: Can we use defer for this?

pkg/kine/server/otel.go
```diff
@@ -0,0 +1,30 @@
...
+	getCnt    metric.Int64Counter
+	listCnt   metric.Int64Counter
+	countCnt  metric.Int64Counter
+	createCnt metric.Int64Counter
+	deleteCnt metric.Int64Counter
+	updateCnt metric.Int64Counter
+)
+
+func init() {
+	listCnt, _ = meter.Int64Counter(fmt.Sprintf("%s.list", name), metric.WithDescription("Number of list requests"))
```
marco6, line 24: Same as in the other package: an error here will cause the code to fail down the line. We should fail here.

As a second thing: could we move each counter closer to where it's used? LIke `listCnt` in the `list.go` file etc etc..

pkg/kine/server/otel.go
```diff
@@ -0,0 +1,30 @@
+package server
+
+import (
+	"fmt"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
+)
+
+const name = "k8s-dqlite"
```
marco6, line 10: can we have a more descriptive name? While not exported, this constant will be available in the whole package and will give very little information to whoever sees it (like in auto-complete)

marco6 (changes requested):
Hi Louise! Thanks for this effort! Here is a first pass on this PR.

pkg/kine/logstructured/sqllog/otel.go
louiseschmidtgen: I'm adding it for completeness. FYI: Otel has this noop implementation which will just do nothing if something isn't defined: https://opentelemetry.io/docs/specs/otel/metrics/noop/#counter-add

louiseschmidtgen:
While I had my cluster running idle I found this outlier:
Our compaction ran for 5m40s instead of our normal 6seconds.
Screenshot from 2024-07-18 10-50-58
Screenshot from 2024-07-18 10-50-27

Consequences for the cluster:
Screenshot from 2024-07-17 16-44-08
Screenshot from 2024-07-17 16-41-07
Screenshot from 2024-07-17 16-39-15

cmd/tracing.go
```diff
@@ -0,0 +1,114 @@
...
+
+	traceExporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithGRPCConn(conn))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create trace exporter: %w", err)
+	}
+	return traceExporter, nil
+}
+
+func newTraceProvider(ctx context.Context, res *resource.Resource) (*trace.TracerProvider, error) {
+	traceExporter, err := newExporter(ctx)
```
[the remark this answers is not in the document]
louiseschmidtgen: What do you mean by 2 connections here? We are not using the same exporter for the metrics since Jaeger can't digest those.

louiseschmidtgen:
The next compaction spike should show which sql query is our culprit or if we have a bug in the kine layer. If it's in the sql query we will get more insight from Marco's work on metrics in dqlite.

louiseschmidtgen:
Side-note: Our kine layer already tracks some metrics here: `http://127.0.0.1:9042/metrics`:
See an example output [here](https://pastebin.canonical.com/p/htssK8WSN2/)

cmd/tracing.go
```diff
@@ -0,0 +1,114 @@
...
+			semconv.ServiceNameKey.String("k8s-dqlite"),
+		),
+	)
+
+	tracerProvider, err := newTraceProvider(ctx, res)
+	if err != nil {
+		handleErr(err)
+		return
+	}
+	shutdownFuncs = append(shutdownFuncs, tracerProvider.Shutdown)
```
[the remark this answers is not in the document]
louiseschmidtgen: It says that shutdown will flush any remaining spans and shut down the exporter.

cmd/tracing.go
```diff
@@ -0,0 +1,114 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// setupOTelSDK bootstraps the OpenTelemetry pipeline.
+// If it does not return an error, make sure to call shutdown for proper cleanup.
+func setupOTelSDK(ctx context.Context) (shutdown func(context.Context) error, err error) {
+	var shutdownFuncs []func(context.Context) error
+
+	shutdown = func(ctx context.Context) error {
```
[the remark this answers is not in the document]
louiseschmidtgen: True this makes sense if you collect logs etc here too.

pkg/kine/logstructured/sqllog/otel.go
louiseschmidtgen: I'm not sure what you mean here? Could you please explain how this would look like?

louiseschmidtgen:
Screenshot from 2024-07-19 09-26-04
In the idle cluster compaction runs at ~6s, every now and then we have a few that take longer, there seems to be a slowdown in these sql queries. Maybe this is due to admissions?

cmd/tracing.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		shutdownFuncs = nil
+		return err
+	}
+
+	handleErr := func(inErr error) {
+		err = errors.Join(inErr, shutdown(ctx))
+	}
+
+	res, err := resource.New(ctx,
```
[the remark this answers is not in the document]
louiseschmidtgen: I will print a message, I don't think we should fail here as the only consequence is that our traces don't have a nice service-name.

cmd/tracing.go
```diff
@@ -0,0 +1,114 @@
...
+	shutdownFuncs = append(shutdownFuncs, meterProvider.Shutdown)
+	otel.SetMeterProvider(meterProvider)
+
+	return
+}
+
+func initConn() (*grpc.ClientConn, error) {
+	// It connects the OpenTelemetry Collector through local gRPC connection.
+	// You may replace `localhost:4317` with your endpoint.
+	conn, err := grpc.NewClient("localhost:4317",
```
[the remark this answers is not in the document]
louiseschmidtgen: I've added a flag `otel-listen`.

pkg/kine/server/update.go
```diff
@@ -29,12 +31,59 @@ func (l *LimitedServer) update(ctx context.Context, rev int64, key string, value
 		ok  bool
 		err error
 	)
+	ctx, updateSpan := tracer.Start(ctx, "update")
```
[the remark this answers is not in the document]
louiseschmidtgen: Here, I moved the child traces into logstructured. I would like to keep one span here to see which logical path it chooses `backend.Create` or `backend.Update`.

pkg/kine/server/update.go
```diff
@@ -29,12 +31,59 @@ func (l *LimitedServer) update(ctx context.Context, rev int64, key string, value
...
+	defer updateSpan.End()
+	updateSpan.SetAttributes(
+		attribute.String("key", key),
+		attribute.Int64("revision", rev),
+		attribute.Int64("lease", lease),
+	)
 
 	if rev == 0 {
+		ctx, span := tracer.Start(ctx, "backend.create")
```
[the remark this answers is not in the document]
louiseschmidtgen: Moved this trace into logstructured.

cmd/tracing.go
```diff
@@ -0,0 +1,114 @@
...
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create gRPC connection to collector: %w", err)
+	}
+
+	return conn, err
+}
+
+func newExporter(ctx context.Context) (trace.SpanExporter, error) {
```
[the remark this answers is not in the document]
louiseschmidtgen: I decided to remove the nested calls and sorted the functions in the order they are now called in `setupOTelSDK`, hoping this will improve the readability.

cmd/tracing.go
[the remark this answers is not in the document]
louiseschmidtgen: I've been trying to get this [example with and otel-collector](https://github.com/open-telemetry/opentelemetry-go/blob/main/example/otel-collector/README.md) to work, but I am unable to get the traces into jaeger via this collector. I kept it as is.

cmd/tracing.go
```diff
@@ -0,0 +1,114 @@
...
+	shutdown = func(ctx context.Context) error {
+		var err error
+		for _, fn := range shutdownFuncs {
+			err = errors.Join(err, fn(ctx))
+		}
+		shutdownFuncs = nil
+		return err
+	}
+
+	handleErr := func(inErr error) {
```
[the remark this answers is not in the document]
louiseschmidtgen: I've removed it alltogether.

cmd/tracing.go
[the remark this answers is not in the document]
louiseschmidtgen: I've created a flag `otel-listen` where you can configure the otel connection.

pkg/kine/server/get.go
louiseschmidtgen: I assigned the error so that I can record it in the span avoiding the duplication.

pkg/kine/server/create.go
louiseschmidtgen: This part is removed.

pkg/kine/server/create.go
louiseschmidtgen: So `span.RecordError(err)` does not stop the span, it will also record just `nil`. Why would you like to use defer here?

marco6 (approved):
This is looking so much better! LGTM! Thanks!

PR #127, opened 2024-07-11
Instrument SQLite

marco6:
This PR enables monitoring SQLite performance metrics. In detail, it adds the following metrics:

 - pages written
   - cache spill (when the write cache needs to be unloaded as the transaction is too big)
 - pages read
   - read cache hits
 - time spent in a read transaction
 - time spent in a write transaction

These will be useful to understand how a new query impacts the database.

### DIFF
--- a/pkg/instrument/extension.c
+++ b/pkg/instrument/extension.c
@@ -1,0 +1,77 @@
+#include "extension.h"
+
+#include <sqlite3.h>
+#include <stdatomic.h>
+#include <stdint.h>
+#include <string.h>
+
+static volatile sqlite3_stats_t global = {0};
+    
+static sqlite3_error collect_profile_stats(unsigned int event, void *pCtx, void *P, void *X) {
+    if (event == SQLITE_TRACE_PROFILE) {
+        sqlite3_stmt* stmt = P;
+        sqlite3* connection = sqlite3_db_handle(stmt);
+
+        int64_t stmt_time_ns = *(sqlite3_int64*)(X);
+        int cache_hit, cache_miss, cache_write, cache_spill, _;
+        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_HIT, &cache_hit, &_, 1);
+        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_MISS, &cache_miss, &_, 1);
+        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_WRITE, &cache_write, &_, 1);
+        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_SPILL, &cache_spill, &_, 1);
+
+        atomic_fetch_add(&global.pages_cache_hit, cache_hit);
+        atomic_fetch_add(&global.pages_cache_miss, cache_miss);
+        atomic_fetch_add(&global.pages_cache_write, cache_write);
+        atomic_fetch_add(&global.pages_cache_spill, cache_spill);
+
+        if (sqlite3_stmt_readonly(stmt)) {
+            atomic_fetch_add(&global.read_txn_time_ns, stmt_time_ns);
+        } else {
+            atomic_fetch_add(&global.write_txn_time_ns, stmt_time_ns);
+        }
+        
+        return SQLITE_OK;
+    }
+    
+    if (event == SQLITE_TRACE_CLOSE) {
+        sqlite3_trace_v2(P, SQLITE_TRACE_PROFILE|SQLITE_TRACE_CLOSE, NULL, NULL);
+    }
+    
+    return SQLITE_OK;
+}
+
+sqlite3_error sqlite3_stats(sqlite3_stats_t* stats, int reset) {
+    if (stats == NULL) {
+        return SQLITE_ERROR;
+    }
+
+    if (reset) {
+        stats->pages_cache_write = atomic_exchange(&global.pages_cache_write, 0);
+        stats->pages_cache_hit = atomic_exchange(&global.pages_cache_hit, 0);
+        stats->pages_cache_miss = atomic_exchange(&global.pages_cache_miss, 0);
+        stats->pages_cache_miss = atomic_exchange(&global.pages_cache_spill, 0);
+        stats->read_txn_time_ns = atomic_exchange(&global.read_txn_time_ns, 0);
+        stats->write_txn_time_ns = atomic_exchange(&global.write_txn_time_ns, 0);
+    } else {
+        stats->pages_cache_write = atomic_load(&global.pages_cache_write);
+        stats->pages_cache_hit = atomic_load(&global.pages_cache_hit);
+        stats->pages_cache_miss = atomic_load(&global.pages_cache_miss);
+        stats->pages_cache_miss = atomic_load(&global.pages_cache_spill);
+        stats->read_txn_time_ns = atomic_load(&global.read_txn_time_ns);
+        stats->write_txn_time_ns = atomic_load(&global.write_txn_time_ns);
+    }
+
+    return SQLITE_OK;
+}
+
+static sqlite3_error auto_instrument_connection(sqlite3 *connection, const char** pzErrMsg, const struct sqlite3_api_routines* pThunk) {
+    return sqlite3_trace_v2(connection, SQLITE_TRACE_PROFILE|SQLITE_TRACE_CLOSE, collect_profile_stats, NULL);
+}
+
+sqlite3_error sqlite3_instrument() {
+    return sqlite3_auto_extension((void (*)())(auto_instrument_connection));
+}
+
+void sqlite3_uninstrument(sqlite3 *connection) {
+    sqlite3_cancel_auto_extension((void (*)())(auto_instrument_connection));
+}
--- a/pkg/instrument/extension.c
+++ b/pkg/instrument/extension.c
@@ -5,64 +5,8 @@
 #include <stdint.h>
 #include <string.h>
 
-static volatile sqlite3_metrics_t global = {0};
+static volatile sqlite3_metrics_t global_metrics = {0};
 
-sqlite3_error sqlite3_collect_metrics(unsigned int event, void *pCtx, void *P, void *X) {
-    if (event == SQLITE_TRACE_PROFILE) {
-        sqlite3_stmt* stmt = P;
-        sqlite3* connection = sqlite3_db_handle(stmt);
-
-        int64_t stmt_time_ns = *(sqlite3_int64*)(X);
-        int cache_hit, cache_miss, cache_write, cache_spill, _;
-        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_HIT, &cache_hit, &_, 1);
-        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_MISS, &cache_miss, &_, 1);
-        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_WRITE, &cache_write, &_, 1);
-        sqlite3_db_status(connection, SQLITE_DBSTATUS_CACHE_SPILL, &cache_spill, &_, 1);
-
-        atomic_fetch_add(&global.pages_cache_hit, cache_hit);
-        atomic_fetch_add(&global.pages_cache_miss, cache_miss);
-        atomic_fetch_add(&global.pages_cache_write, cache_write);
-        atomic_fetch_add(&global.pages_cache_spill, cache_spill);
-
-        if (sqlite3_stmt_readonly(stmt)) {
-            atomic_fetch_add(&global.read_txn_time_ns, stmt_time_ns);
-        } else {
-            atomic_fetch_add(&global.write_txn_time_ns, stmt_time_ns);
-        }
-        
-        return SQLITE_OK;
-    }
-    
-    if (event == SQLITE_TRACE_CLOSE) {
-        sqlite3_trace_v2(P, SQLITE_TRACE_PROFILE|SQLITE_TRACE_CLOSE, NULL, NULL);
-    }
-    
-    return SQLITE_OK;
-}
-
-sqlite3_error sqlite3_metrics(sqlite3_metrics_t* stats, int reset) {
-    if (stats == NULL) {
-        return SQLITE_ERROR;
-    }
-
-    if (reset) {
-        stats->pages_cache_write = atomic_exchange(&global.pages_cache_write, 0);
-        stats->pages_cache_hit = atomic_exchange(&global.pages_cache_hit, 0);
-        stats->pages_cache_miss = atomic_exchange(&global.pages_cache_miss, 0);
-        stats->pages_cache_miss = atomic_exchange(&global.pages_cache_spill, 0);
-        stats->read_txn_time_ns = atomic_exchange(&global.read_txn_time_ns, 0);
-        stats->write_txn_time_ns = atomic_exchange(&global.write_txn_time_ns, 0);
-    } else {
-        stats->pages_cache_write = atomic_load(&global.pages_cache_write);
-        stats->pages_cache_hit = atomic_load(&global.pages_cache_hit);
-        stats->pages_cache_miss = atomic_load(&global.pages_cache_miss);
-        stats->pages_cache_miss = atomic_load(&global.pages_cache_spill);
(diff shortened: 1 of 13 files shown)


marco6:
Depends on #129 and #130

marco6:
Dropping dependency from #129  since it was merged.

test/util_test.go
```diff
@@ -92,8 +147,83 @@ func startDqlite(ctx context.Context, tb testing.TB) endpoint.Config {
...
+	UNION ALL
+
+	SELECT id + 1
+	FROM gen_id
+	WHERE id < ?
+), revision AS(
+	SELECT COALESCE(MAX(id), 0) AS base
+	FROM kine
+)
+INSERT INTO kine(
```
TheSignPainter98, line 181: This may be an existing convention, but I think it could be nice to put some newlines between the separate parts of these queries

test/util_test.go
```diff
@@ -92,8 +147,83 @@ func startDqlite(ctx context.Context, tb testing.TB) endpoint.Config {
...
+FROM kine AS kv
+JOIN (
+	SELECT MAX(mkv.id) as id
+	FROM kine mkv
+	WHERE  ?||'/' <= mkv.name AND mkv.name < ?||'0'
+	GROUP BY mkv.name
+) maxkv ON maxkv.id = kv.id
+WHERE kv.deleted = 0
+ORDER BY kv.name
+LIMIT %d`, numUpdates)
```
TheSignPainter98, line 204: I think `numUpdates` slightly gets lost at the end of this line as the queries spans multiple lines. Perhaps this could be on a newline?

pkg/instrument/extension.h
```diff
@@ -0,0 +1,41 @@
...
+
+// sqlite3_instrument register an auto extension that 
+// instruments all new connections to collect key sqlite
+// performance metrics. Existing connections will not be 
+// affected. This call uses the trace hook to collect data.
+// Replacing the hook will remove instrumentation for the
+// connection. In this case, it is possible to call 
+// [sqlite3_collect_metrics] to keep the instrumentation
+// active in a custom trace hook.
+sqlite3_error sqlite3_instrument();
```
TheSignPainter98, line 24: `s/register/registers/`
`s/an auto extension that/an auto extension which/`

What does 'key' sqlite performance metrics mean here? Is it key-performance metrics or 'key' (as in 'important') sqlite metrics. If it's the latter, I think 'key' is very unspecific so should be dropped.

There's some slight inconsistency here, the doc comments for `sqlite3_instrument` and `_uninstrument` mention different hooks, could these be made the same? Also, do we need to mention which hooks are set? Could it be kept as an implementation detail, or could this cause unexpected clashes?

pkg/instrument/extension.h
```diff
@@ -0,0 +1,41 @@
...
+// It will not restore previous commit or trace hook.
+// If an error occurs, the conneciton is left unchanged
+// and the instrumentation context is not freed.
+void sqlite3_uninstrument();
+
+// sqlite3_stats reads copies performance metrics into stats.
+// It returns SQLITE_ERROR error if stats is null, SQLITE_OK
+// otherwise.
+// If reset is != 0, it resets each metric to 0.
+sqlite3_error sqlite3_metrics(sqlite3_metrics_t* stats, int reset);
```
TheSignPainter98, line 37: `s/reads//`
`s/copies performance metrics/copies the current performance metrics/`?

Also I've said elsewhere (if the comment will appear though) but the `reset` parameter feels odd here

test/compaction_test.go
```diff
@@ -0,0 +1,101 @@
...
+			defer cancel()
+
+			kine := newKine(ctx, b, &kineOptions{
+				backendType: backendType,
+				setup: func(db *sql.DB) error {
+					// Make sure there's enough rows deleted to have
+					// b.N rows to compact.
+					delCount := b.N + sqllog.SupersededCount
+
+					// Also, make sure there's some uncollectable data, so
```
TheSignPainter98, line 85: This is more pedantic but this should also be 'there are' as 'data' is plural ;)

TheSignPainter98 (changes requested):
Great work, Marco! Here are a some things I think we should consider

Also, some of my comments will have disappeared from the code view due to pushes, but I think most of the notes there should still apply, so please be sure to find them! :D

pkg/instrument/extension.c
```diff
@@ -0,0 +1,77 @@
+#include "extension.h"
+
+#include <sqlite3.h>
+#include <stdatomic.h>
+#include <stdint.h>
+#include <string.h>
+
+static volatile sqlite3_stats_t global = {0};
```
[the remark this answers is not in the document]
marco6: I'm going for `global_metrics` after I renamed all to `metrics` to follow Go convention.

pkg/instrument/extension.c
```diff
@@ -0,0 +1,77 @@
...
+    
+    if (event == SQLITE_TRACE_CLOSE) {
+        sqlite3_trace_v2(P, SQLITE_TRACE_PROFILE|SQLITE_TRACE_CLOSE, NULL, NULL);
+    }
+    
+    return SQLITE_OK;
+}
+
+sqlite3_error sqlite3_stats(sqlite3_stats_t* stats, int reset) {
+    if (stats == NULL) {
```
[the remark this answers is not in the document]
marco6: Done. I kept the error return to make sure that if we add stuff in the future the interface doesn't change.

pkg/instrument/extension.h
```diff
@@ -0,0 +1,32 @@
...
+} sqlite3_stats_t;
+
+// sqlite3_instrument instruments the connection represented
+// by the first argument so it can take measurements of key
+// sqlite performance metrics.
+// It will replace both the commit and the trace hook for 
+// the connection. If an error occurs, the connection is 
+// not touched (i.e. neither the trace nor the commit hook
+// are replaced).
+sqlite3_error sqlite3_instrument();
```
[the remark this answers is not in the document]
marco6: This needs one more round as I changed all doc comments. This one in particular was coming from a previous version that took a `sqlite3* connection` as an argument. Please review the updated comments instead.

pkg/instrument/extension.h
```diff
@@ -0,0 +1,32 @@
...
+
+// sqlite3_uninstrument remove instrumentation connection
+// and free the associated context.
+// It will not restore previous commit or trace hook.
+// If an error occurs, the conneciton is left unchanged
+// and the instrumentation context is not freed.
+void sqlite3_uninstrument();
```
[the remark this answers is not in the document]
marco6: I changed the name to `sqlite3_deinstrument`, but could you review the new doc comment?

pkg/instrument/sqlite.go
```diff
@@ -0,0 +1,50 @@
...
+	_ "unsafe"
+
+	"github.com/mattn/go-sqlite3"
+)
+
+// #cgo LDFLAGS: -lsqlite3
+// #include "extension.h"
+import "C"
+
+func SQLite() (*SQLiteMonitor, error) {
```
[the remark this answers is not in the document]
marco6: They are now called `{Start,Stop}SQLiteMonitoring`, to follow Go naming.

test/util_test.go
```diff
@@ -19,39 +23,82 @@ func init() {
...
 
 	if !strings.Contains(endpointConfig.Endpoint, "?") {
 		endpointConfig.Endpoint += "?"
 	}
-	for _, v := range qs {
+	for _, v := range options.endpointParameters {
 		endpointConfig.Endpoint = fmt.Sprintf("%s&%s", endpointConfig.Endpoint, v)
 	}
-	config, backend, err := endpoint.ListenAndReturnBackend(ctx, endpointConfig)
+	config, backend, err := endpoint.ListenAndReturnBackend(ctx, *endpointConfig)
```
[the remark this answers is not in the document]
marco6: This interface is part of upstream and I'm not sure I should be changing it in this PR. Major refactor is indeed needed in this project, so I don't disagree - I just think this is not the right PR.

test/util_test.go
marco6: It is an existing convention not to.

pkg/instrument/extension.h
marco6: We need to keep the description of the hooks as only one hook of each kind can be set up (i.e. if someone does, he will remove instrumentation from that connection).

Regarding the 'key` it's as in 'important metrics'. What do you suggest instead? Just dropping `key`?

pkg/instrument/extension.h
marco6: As said above, we need this reset to be atomic or we'd loose some data.

marco6:
Some of the comments of the review, and the related changes are actually referring to #130. Given that the review is in this PR though, I think they should be kept here. 

If someone disagrees, I will cherry pick them and move them over.

pkg/instrument/extension.h
TheSignPainter98: I think dropping the word 'key' would be cleanest

test/util_test.go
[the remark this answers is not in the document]
TheSignPainter98: Okay, putting this out of the scope of this PR sounds reasonable to me

pkg/instrument/extension.c
```diff
@@ -0,0 +1,78 @@
...
+}
+
+sqlite3_error sqlite3_collect_metrics(unsigned int event, void *pCtx, void *P, void *X) {
+    sqlite3_stmt* stmt;
+    sqlite3* connection;
+
+    int64_t stmt_time_ns;
+    int cache_hit, cache_miss, cache_write, cache_spill, _;
+    
+    switch (event) {
```
TheSignPainter98, line 50: Not sure if this is an existing convention, but would it be better to move these into the case where they're used?

pkg/instrument/extension.h
```diff
@@ -0,0 +1,39 @@
...
+// In this case, it is possible to call [sqlite3_collect_metrics]
+// to keep the instrumentation active in a custom trace hook.
+sqlite3_error sqlite3_instrument();
+
+// sqlite3_deinstrument removes instrumentation connection
+// and free the associated context.
+// It will not restore previous commit or trace hook.
+// If an error occurs, the conneciton is left unchanged
+// and the instrumentation context is not freed.
+void sqlite3_deinstrument();
```
TheSignPainter98, line 31: I think to match the previous function, we should mention removing instrumentation from 'existing connections, freeing their associated context.' I'm not sure what the associated context is, I think it'd be better to clarify or drop this.

In the second sentence, the 'it' s not immediately clear, perhaps 'Calling this function'?

In the 'If an error occurs' paragraph, would 'existing connections are left unchanged work'? (Also there's a typo on connection :P)

pkg/instrument/extension.h
```diff
@@ -0,0 +1,39 @@
...
+
+// sqlite3_metrics copies the current performance metrics into its first argument.
+// If reset is != 0, it resets each metric to 0.
+sqlite3_error sqlite3_metrics(sqlite3_metrics_t* metrics, int reset);
```
TheSignPainter98, line 35: I think resetting to `0` seems like an implementation detail, perhaps 'it resets the metrics' could be sufficient?

pkg/instrument/extension.h
```diff
@@ -0,0 +1,39 @@
...
+
+// sqlite3_collect_metrics is a trace hook that collects performance
+// metrics during the SQLITE_TRACE_PROFILE event.
+sqlite3_error sqlite3_collect_metrics(unsigned int event, void *pCtx, void *P, void *X);
```
TheSignPainter98, line 39: I don't think this will be used as a trace hook, but rather as a call to be made within someone's custom trace hook. Perhaps something like this could work?
```
// sqlite3_collect_metrics can be called to collect performance
// metrics during a SQLITE_TRACE_PROFILE event.

File: pkg/instrument/sqlite.go [hunk 16]
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/mattn/go-sqlite3"
+)
+
+// #cgo LDFLAGS: -lsqlite3
+// #include "extension.h"
+import "C"
+
+// StartSQLiteMonitoring instruments sqlite to collect metrics
+// on all connection opened after the call (i.e. existing connections
+// will not be instrumented).
```
TheSignPainter98, line 16: I think this can be simplified to 'all subsequent connections'

pkg/instrument/sqlite.go
```diff
@@ -0,0 +1,68 @@
...
+func StartSQLiteMonitoring() error {
+	if err := C.sqlite3_instrument(); err != sqlite3.SQLITE_OK {
+		return sqlite3.ErrNo(err)
+	}
+	return nil
+}
+
+// StopSQLiteMonitoring removes instrumentation from sqlite.
+// Existing connections will still be insturmented until they
+// are closed.
```
TheSignPainter98, line 27: I think we can be more flattering to ourselves here ;) Perhaps the logic could be reversed here?
```
// After this call, no new connections are instrumented.
```

pkg/instrument/sqlite.go
```diff
@@ -0,0 +1,68 @@
...
+	PagesCacheHit   uint64
+	PagesCacheMiss  uint64
+	PagesCacheSpill uint64
+
+	TransactionReadTime  time.Duration
+	TransactionWriteTime time.Duration
+}
+
+// ResetSQLiteMetrics returns the current metrics and
+// resets their value to 0.
```
TheSignPainter98, line 43: I think the model is slightly different to the C-side here, so there isn't anything external being reset to zero just here. Perhaps this function could be renamed? Perhaps something like `FinalizeSQLiteMetrics` (with the US spelling of 'finalise' :P)? I'm really not sure what a better name would be here

pkg/instrument/extension.c
marco6: In C, you cannot declare variables inside a switch case. The only way would be to create a scope for that, but it is a bit weird? like

```C
switch (event) {
case SQLITE_TRACE_PROFILE:
    {
        ...
    }
    break;
 ...
 ```

File: pkg/instrument/extension.h (hunk 13 above)
Comment (marco6) [reply pkg/instrument/extension.h:31]: I think this is coming again from a previous version. I simplified with:

```
// sqlite3_deinstrument stops instrumenting new connections.
// Existing connections are left unchanged.
```

File: pkg/instrument/sqlite.go (hunk 18 above)
Comment (marco6) [reply pkg/instrument/sqlite.go:43]: This will affect subsequent calls though, so the reset is an effect which is visible from the outside and it should be valid. Moreover, while the *current* usage is clearly using it just to reset (and discards the result) that might not be the case during live server executions as you would use the result of this function. In that case wouldn't it be weird if I would take data multiple times from a `FinalizeXXX` function?

File: pkg/instrument/extension.c (hunk 12 above)
Comment (TheSignPainter98) [reply pkg/instrument/extension.c:50]: Ah of course, I forgot about this! If there's an existing convention, I think let's go with that, but I think if in future other interesting cases are added to this `switch`, it would be messy to have all of the variables declared above (especially if some names end up being shared!), so although somewhat ugly, creating a scope here would feel cleaner to me

Comment (marco6) [reply pkg/instrument/extension.c]: I brought back `if`s as they seem to be more readable.

**Review by bschimke95 — APPROVED** [review]
LGTM, minor comment

File: pkg/instrument/extension.c [hunk 19]
```diff
@@ -0,0 +1,75 @@
...
+        metrics->pages_cache_miss = atomic_load(&global_metrics.pages_cache_miss);
+        metrics->pages_cache_miss = atomic_load(&global_metrics.pages_cache_spill);
+        metrics->read_txn_time_ns = atomic_load(&global_metrics.read_txn_time_ns);
+        metrics->write_txn_time_ns = atomic_load(&global_metrics.write_txn_time_ns);
+    }
+
+    return SQLITE_OK;
+}
+
+sqlite3_error sqlite3_collect_metrics(unsigned int event, void *pCtx, void *P, void *X) {
```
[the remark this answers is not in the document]
TheSignPainter98: Just for the record of our conversation on MM, in the absence of any strong candidates for better names, we believe that using `P` and `X`, [as introduced by sqlite itself](https://sqlite.org/c3ref/c_trace.html), is okay here

TheSignPainter98 (approved):
Looks nice and clean to me, thanks for the changes! :D

PR #129, opened 2024-07-12
Update go-sqlite3 to the last version

marco6:
The latest stable version of `go-sqlite` is `v1.14` or later, not `v2`, as described [in the readme](https://github.com/mattn/go-sqlite3/blob/master/README.md), and AFAIK `v2` has been retired (see the `+incompatible` tag in the go mod).

This PR points `go-sqlite` dependency to `v1.14`.


bschimke95 (approved):
LGTM

PR #131, opened 2024-07-12
Instrument dqlite

marco6:
This PR adds instrumentation of dqlite to obtain 2 dqlite-related metrics: the amount of bytes read/written by the client to the server.

This metric is important as it might me proxied by a connection to a follower to the leader, clogging even more the network.

All metrics gathered from sqlite are still a valid indicator for dqlite.


marco6:
This PR depends on a [very small fix](https://github.com/canonical/go-dqlite/pull/304), but the result is that we are depending on something that is not released yet.

Maybe we should wait for a new release.

marco6:
To summarize what we discussed on Friday: we cannot merge any PR which relies on an unreleased version of any dependency. As such I'm marking this as draft, waiting for the dqlite team to release a new version with the fix.

test/util_test.go
```diff
@@ -152,10 +159,19 @@ func (ks *kineServer) ReportMetrics(b *testing.B) {
 	b.ReportMetric(float64(sqliteMetrics.PageCacheWrites)/float64(b.N), "page-writes/op")
 	b.ReportMetric(float64(sqliteMetrics.TransactionReadTime)/float64(time.Second)/float64(b.N), "sec-reading/op")
 	b.ReportMetric(float64(sqliteMetrics.TransactionWriteTime)/float64(time.Second)/float64(b.N), "sec-writing/op")
+
+	if ks.dqliteListener != nil {
+		dqliteMetrics := ks.dqliteListener.Metrics()
+		b.ReportMetric(float64(dqliteMetrics.BytesRead)/float64(b.N), "network-reads/op")
+		b.ReportMetric(float64(dqliteMetrics.BytesWritten)/float64(b.N), "network-writes/op")
```
TheSignPainter98, line 166: I think the metric names are slightly off here, it looks like the number of times the network is used, but the value is in terms of bytes. Perhaps these could be renamed to something like these?
- `network-bytes-read/op`
- `network-bytes-written/op`

TheSignPainter98 (changes requested):
Looking good, here's a first pass. There were a couple of places where I wasn't quite clear on how some parts fitted together (particularly `Listen` and `Connect`), but I this can be remedied with some docs

pkg/instrument/dqlite.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil, fmt.Errorf("can't setup probes")
+}
+
+func (l *Listener) run() {
+	defer close(l.AcceptChan)
+	for {
+		conn, err := l.listener.Accept()
+		if err != nil {
+			break
+		}
```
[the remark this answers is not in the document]
marco6: This does not ignore errors! It *stops* when an error occurs. Making it possible to retrieve it seems like a good thing, I'll add it. I'm not 100% sure which error should be ignored (i.e. what error is returned when the listener gets closed), but I will find out.

pkg/instrument/dqlite.go
```diff
@@ -0,0 +1,111 @@
...
+	l.wg.Wait()
+	return nil
+}
+
+type fileConn interface {
+	net.Conn
+	File() (*os.File, error)
+}
+
+type wrappedConn struct {
```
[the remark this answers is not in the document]
marco6: I'll go with `metredConn`

pkg/instrument/dqlite.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if conn, ok := conn.(fileConn); ok {
+		return &wrappedConn{conn, l}, nil
+	}
+	return nil, fmt.Errorf("can't setup probes")
+}
+
+func (l *Listener) run() {
```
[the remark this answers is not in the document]
marco6: As discussed on MM we went for `NewListener` with a `Load` method, so that we split the two phases.

marco6:
Marking as ready as both dependencies were merged.

pkg/instrument/dqlite.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"net"
+	"os"
+	"sync"
+	"sync/atomic"
+)
+
+type Listener struct {
+	Network, Address        string
+	AcceptedChan            chan net.Conn
```
TheSignPainter98, line 15: Do we need the `Chan` suffix? I think it should be clear from usage that this is a channel. Would `AcceptedConns` be better? There's a line later which this change would make read---
```go
l.AcceptedConns <- conn
```
which feels nice

pkg/instrument/dqlite.go
```diff
@@ -0,0 +1,125 @@
...
+	BytesWritten int64
+	BytesRead    int64
+}
+
+func NewListener(network, address string) *Listener {
+	return &Listener{
+		Network:      network,
+		Address:      address,
+		AcceptedChan: make(chan net.Conn),
+		netListener:  nil,
```
TheSignPainter98, line 32: As other fields are taking their zero values, I think it's okay to omit this `nil`-assignment here. I think the absence of `netListener` in `NewListener` is a sufficient indicator that this field is nillable

pkg/instrument/dqlite.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func (l *Listener) Listen(ctx context.Context) error {
+	if l.netListener != nil {
+		return fmt.Errorf("listener already running")
+	}
+	var lc net.ListenConfig
+	netListener, err := lc.Listen(ctx, l.Network, l.Address)
```
TheSignPainter98, line 41: A good convention to keep to is to use `var v T` declarations when `v` is later assigned to before being read. I think it would cleaner to explicitly assign this to its zero value to indicate that it will be used in that state

Also the name `lc` is quite short, could this be made more descriptive, perhaps as `netListenerConfig`? I'm on the fence about this one

pkg/instrument/dqlite.go
```diff
@@ -0,0 +1,125 @@
...
+func (l *Listener) Connect(ctx context.Context, _ string) (net.Conn, error) {
+	dialer := &net.Dialer{}
+	conn, err := dialer.DialContext(ctx, l.Network, l.Address)
+	if err != nil {
+		return nil, err
+	}
+	if conn, ok := conn.(fileConn); ok {
+		return &metredConn{conn, l}, nil
+	}
+	return nil, fmt.Errorf("can't setup probes")
```
TheSignPainter98, line 74: I think usually we avoid contractions in error messages so we should `s/can't/cannot/` here

TheSignPainter98 (changes requested):
This is looking really nice, just a couple of nitpicks then I think this is ready

pkg/instrument/dqlite.go
marco6: TBH I copied this snippet [from the Go runtime](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/dial.go;l=806), so I don't really have any opinion on this.

I'll go with both suggestions.

pkg/instrument/dqlite.go
marco6: It's about the 100th time you correct me there lol

TheSignPainter98 (approved):
This is looking really nice now, thanks for the changes, Marco!

PR #132, opened 2024-07-16
Automate prepared statements

marco6:
This PR aims at reducing the boilerplate associated with preparing statements in a `Dialect`.


marco6:
Depends on canonical/go-dqlite#308 .

marco6:
Marking as ready since dqlite released the fix.

bschimke95 (approved):
LGTM

marco6:
On my machine this change makes the benchmarks unstable because of canonical/dqlite#675 .

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+	"errors"
+	"sync"
+	"sync/atomic"
+)
+
+type DB struct {
+	underlying *sql.DB
+	mu         sync.Mutex
+	maxSize    int
+	store      map[string]*lruEntry
```
TheSignPainter98, line 15: It's not immediately clear to me what this is a store of, also if we add more similar stores in future we'll have a name-clash. Could this be renamed something like `preparedQueryCache`?

Also, this cache functionality, although strongly related to the `DB`, I wonder whether it ought to go in its own separate struct? I think it could be really nice and clean to have calls like `db.cache.get()` and `db.cache.put()`

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+	db.lruList.add(entry)
+}
+
+func (db *DB) drop(entry *lruEntry) error {
+	db.lruList.remove(entry)
+	delete(db.store, entry.query)
+	return entry.stmt.Close()
+}
+
+type stmt struct {
```
TheSignPainter98, line 128: The name `stmt` clashes with other `stmt` variables above. Is it possible rename this to something like `preparedStmt`? I see this getting returned from methods like `prepare` above

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+	*sql.Stmt
+	count atomic.Int32
```
TheSignPainter98, line 130: It's not immediately clear what this is a count of. If this is a reference count, I think it would be clearer to rename this field to something like `refCount`

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (s *stmt) Ref() *stmt {
+	s.count.Add(1)
+	return s
+}
```
TheSignPainter98, line 136: I don't think the name `Ref` strongly conveys the meaning of this method, in fact when I first saw it above, I had no idea what it was doing. As this is incrementing the reference count, could this method be renamed to something like `IncrementRefCount`, `IncrRefCount` or perhaps even `CountRef`? (Not sure about the last one)

Also, this function returning its receiver seems quite strange as the caller will already be in possession of a copy of the receiver (of course, if we were using a certain language with move semantics, I'd likely say something different). If we wish to keep this behaviour, I think the name of this function should be clarified, perhaps to something like `NewRef` or `MakeRef`?

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+	return stms.QueryContext(ctx, args...)
+}
+
+func (db *DB) Close() error {
+	db.mu.Lock()
+	defer db.mu.Unlock()
+
+	errs := []error{}
+	for len(db.store) != 0 {
+		if err := db.drop(db.lruList.tail()); err != nil {
```
TheSignPainter98, line 51: The conditions here seem a little muddled in my mind---we're iterating whilst the store is non-empty, getting the tail of the list each time, which is implicitly changed by calling `db.drop`. Do we need to iterate over the cached queries in a particular order? I think it would be cleaner to do
```go
for _, cacheEntry := range db.store {
    ...
}
```
If the order matters and we do indeed wish to drop the most used one last, I think could be nice to have an iterator over the LRU list like this:
```go
func (ll *lruList) IterateFromLeast() *lruFromLeastIterator {
    return &lruIterator{
        curr: ll.tail()
    }
}

struct lruFromLeastIterator {
    curr *lruEntry
}

func (it *lruFromLeastIterator) Next(entry **lruEntry) bool {
    if it.curr == nil {
        return false
    }
    *entry = it.curr
    it.curr = it.curr.prev
    return true
}

iter := db.lruList.IterateFromLeast()
var entry *lruEntry
for iter.Next(&entry) {
    if err := db.drop(entry); err != nil {
        errs = append(errs, err)
    }
}

File: pkg/kine/prepared/db.go [hunk 6]
```diff
@@ -0,0 +1,191 @@
...
+	entry, ok := db.store[query]
+	if ok {
+		db.touch(entry)
+		entry.stmt.Close()
+		entry.stmt = stmt.Ref()
+		return
+	}
+
+	if len(db.store) >= db.maxSize {
+		db.drop(db.lruList.tail())
```
TheSignPainter98, line 111: I think if the cache were moved into its own struct, we could add our own method on it to make it clear what we're getting here (somehow the 'tail' of the LRU list being the least recently used and not the most recently used seems like an implementation detail). I think it would be really nice if this line read something like
```go
db.drop(db.cache.LeastRecentlyUsed())
```
Also, I think there's either some missing error propagation here, or, if we're ignoring errors at this point (which seems fairly reasonable), it would be good to explicitly state that we're ignoring them, e.g. with
 ```go
_ = db.drop(...)
```
I actually wonder whether this is the right place to ignore these errors, or whether they should be passed upwards to be ignored elsewhere?

File: pkg/kine/prepared/db.go [hunk 7]
```diff
@@ -0,0 +1,191 @@
...
+func (db *DB) prepare(ctx context.Context, query string) (*stmt, error) {
+	if stmt := db.get(query); stmt != nil {
+		return stmt, nil
+	}
+	prepared, err := db.underlying.PrepareContext(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	result := &stmt{prepared, atomic.Int32{}}
+	result.Ref()
```
TheSignPainter98, line 77: The need to immediately add one to the reference counter is quite fragile, I think in this case, it would be better to have a function which does the value construction for us, perhaps `newStmt(prepared)`?

It also looks very suspicious that the result of `.Ref()` is being ignored, even if it's okay in this case. Also, I think this call to increment the reference counter (if kept) should go immediately above the `return result` line as `db.put` already increments it internally so this would make it clearer what the refCount++ is referring to

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	db.mu.Lock()
+	defer db.mu.Unlock()
+
+	entry, ok := db.store[query]
+	if ok {
+		db.touch(entry)
+		entry.stmt.Close()
+		entry.stmt = stmt.Ref()
```
TheSignPainter98, line 106: The need to close wasn't immediately clear to me, perhaps a comment could help here? I'm thinking something like this perhaps?
```go
// Replace existing entry.
```

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (s *stmt) Close() error {
+	if s.count.Add(-1) <= 0 {
+		return s.Stmt.Close()
+	}
+	return nil
+}
+
+type lruList struct {
+	// lruList is implemented as a circular linked list
```
TheSignPainter98, line 146: Full stop required :)

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+	head *lruEntry
+}
+
+type lruEntry struct {
+	query string
+	stmt  *stmt
+
+	// Adjacent entries, unless at the tail of the list, the
+	// next entry was added less recently tham the current.
+	// Likewise prev was added more recently.
```
TheSignPainter98, line 156: I think the leading 'Adjacent entries' can be dropped or turned into a sort of mini-title by adding a colon ('Adjacent entries:')

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (ll *lruList) add(entry *lruEntry) {
+	if ll.head != nil {
+		entry.prev = ll.head.prev
+		entry.next = ll.head
+		ll.head.prev.next = entry
+		ll.head.prev = entry
+	} else {
+		entry.prev = entry
+		entry.next = entry
```
TheSignPainter98, line 181: As this case will be run first, I think it could be nice to swap the order of these branches

TheSignPainter98 (changes requested):
Great work Marco, here are a few things I think we should consider before merging

pkg/kine/prepared/db.go
marco6: As discussed on MM, I will remove the LRU part.

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,191 @@
...
+	"sync"
+	"sync/atomic"
+)
+
+type DB struct {
+	underlying *sql.DB
+	mu         sync.Mutex
+	maxSize    int
+	store      map[string]*lruEntry
+	lruList    lruList
```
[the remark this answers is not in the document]
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/prepared/db.go
marco6: Removed.

pkg/kine/drivers/generic/generic.go
```diff
@@ -131,37 +131,27 @@ type ErrCode func(error) string
...
+	CountCurrentSQL       string
+	CountRevisionSQL      string
+	AfterSQLPrefix        string
+	AfterSQL              string
+	DeleteSQL             string
+	UpdateCompactSQL      string
+	InsertSQL             string
+	FillSQL               string
+	InsertLastInsertIDSQL string
+	GetSizeSQL            string
```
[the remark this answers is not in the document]
TheSignPainter98: Ah thanks for the context, sounds good to me

pkg/kine/drivers/generic/generic.go
```diff
@@ -394,41 +314,59 @@ func (d *Generic) query(ctx context.Context, txName, sql string, args ...interfa
 		recordOpResult(txName, err, start)
 	}()
 
-	strippedSQL := Stripped(sql)
+	strippedQuery := Stripped(query)
```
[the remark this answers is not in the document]
TheSignPainter98: Woah I missed that! I don't think this type name indicates the laziness, if possible I think renaming this to `LazilyStripped` could be cleaner?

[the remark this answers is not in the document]
marco6: As discussed on MM, while it is true that the aim of that conversion is not clear, it is also not part of this PR. We will need to think if it makes sense or not though.

TheSignPainter98 (changes requested):
Nearly there I think, just a couple of trivials

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -73,13 +73,15 @@ type Dialect interface {
 	GetSize(ctx context.Context) (int64, error)
 	GetCompactInterval() time.Duration
 	GetPollInterval() time.Duration
-	Close()
+	Close() error
```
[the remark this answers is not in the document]
marco6: In general, most of the `Close` return an error, but it's still fine to defer them, depending on the application. For example an `os.File`'s closing can be deferred if you are opening it for reading only (as you don't really care if you fail there). Same with `Rollback`s as if the connection fails (which is the only reason a rollback can fail on most databases), then you don't really care as a closed connection without a commit will make the database rollback.

So, I think it's up to the user. In this case we do check the error though.

TheSignPainter98 (approved):
Excellent, thank you for the improvements!

PR #136, opened 2024-07-22
Improve list test and resize admission test

marco6:
This PR fixes the list test so that it is possible to get reliable metrics from it and adds a test for pagination.

### DIFF
--- a/test/admission_test.go
+++ b/test/admission_test.go
@@ -47,7 +47,7 @@ func TestAdmissionControl(t *testing.T) {
 			var numSuccessfulWriterTxn = atomic.Uint64{}
 			var numSuccessfulReaderTxn = atomic.Uint64{}
 
-			reader := func(first int, last int) {
+			read := func(first int, last int) {
 				defer wg.Done()
 				for i := first; i < last; i++ {
 					key := fmt.Sprintf("Key/%d", i+1)
@@ -58,7 +58,7 @@ func TestAdmissionControl(t *testing.T) {
 				}
 			}
 
-			writer := func(first int, last int) {
+			write := func(first int, last int) {
 				defer wg.Done()
 				for i := first; i < last; i++ {
 					key := fmt.Sprintf("Key/%d", i+1)
@@ -91,10 +91,10 @@ func TestAdmissionControl(t *testing.T) {
 
 			start := time.Now()
 			for i := 0; i < readers; i++ {
-				go reader(i*read_entries, (i+1)*read_entries)
+				go read(i*read_entries, (i+1)*read_entries)
 			}
 			for i := 0; i < writers; i++ {
-				go writer(i*write_entries, (i+1)*write_entries)
+				go write(i*write_entries, (i+1)*write_entries)
 			}
 
 			wg.Wait()
--- a/test/admission_test.go
+++ b/test/admission_test.go
@@ -47,9 +47,9 @@ func TestAdmissionControl(t *testing.T) {
 			var numSuccessfulWriterTxn = atomic.Uint64{}
 			var numSuccessfulReaderTxn = atomic.Uint64{}
 
-			read := func(first int, last int) {
+			read := func(firstKeyNum, lastKeyNum int) {
 				defer wg.Done()
-				for i := first; i < last; i++ {
+				for i := firstKeyNum; i < lastKeyNum; i++ {
 					key := fmt.Sprintf("Key/%d", i+1)
 					_, err := kine.client.Get(ctx, key, clientv3.WithRange(""))
 					if err == nil {
@@ -58,9 +58,9 @@ func TestAdmissionControl(t *testing.T) {
 				}
 			}
 
-			write := func(first int, last int) {
+			write := func(firstKeyNum, lastKeyNum int) {
 				defer wg.Done()
-				for i := first; i < last; i++ {
+				for i := firstKeyNum; i < lastKeyNum; i++ {
 					key := fmt.Sprintf("Key/%d", i+1)
 					new_value := fmt.Sprintf("New-Value-%d", i+1)
 					resp, err := kine.client.Get(ctx, key, clientv3.WithRange(""))
@@ -87,6 +87,7 @@ func TestAdmissionControl(t *testing.T) {
 
 			writers := writeLimit * 4
 			write_entries := 20
+
 			wg.Add(readers + writers)
 
 			start := time.Now()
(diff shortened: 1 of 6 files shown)


marco6:
Depends on #131

marco6:
This PR makes the test run with a significantly higher number of rows. As such some queries in the background remain active quite often at the end of the test (compaction, tll, watchers). 

This makes dqlite segfault quite often. It should be solved with canonical/dqlite#675 which now this PR depends on.

marco6:
Marking as ready since #131 and canonical/dqlite#675 were both merged.

louiseschmidtgen (commented):
Great work Marco! 

One little knit could we improve the readability of the benchmark output by explaining the meaning of `-100` and `-4`? Maybe something like `List/sqlite-100payloadSize`
```
List/sqlite-100/all-4  
List/sqlite-100/pagination-4  
```

test/list_test.go
```diff
@@ -178,32 +178,100 @@ func TestList(t *testing.T) {
...
+				b.Run("pagination", func(b *testing.B) {
+					b.StopTimer()
+					g := NewWithT(b)
+
+					ctx, cancel := context.WithCancel(context.Background())
+					defer cancel()
+
+					kine := newKineServer(ctx, b, &kineOptions{
+						backendType: backendType,
+						setup: func(ctx context.Context, tx *sql.Tx) error {
```
TheSignPainter98, line 248: Not sure if this is something we can change in this PR, but it's not immediately clear what this is setting up for me, it looks like it's setting up a sql channel, could this be renamed? I'm not quite sure what the name here would be. Otherwise, this is okay

test/list_test.go
```diff
@@ -178,32 +178,100 @@ func TestList(t *testing.T) {
...
+							return setup(ctx, tx, payload.size, b.N)
+						},
+					})
+
+					kine.ResetMetrics()
+					b.StartTimer()
+					nextKey := "key/"
```
TheSignPainter98, line 255: The logic around `nextKey` is very strange to me, could this be simplified somehow? Also, it looks like the initial `key/` gets lost, is that okay?

TheSignPainter98 (changes requested):
Looking good, just a few trivials

test/list_test.go
```diff
@@ -178,32 +178,88 @@ func TestList(t *testing.T) {
...
 func BenchmarkList(b *testing.B) {
-	for _, backendType := range []string{endpoint.SQLiteBackend, endpoint.DQLiteBackend} {
-		b.Run(backendType, func(b *testing.B) {
-			b.StopTimer()
-			g := NewWithT(b)
+	setup := func(ctx context.Context, tx *sql.Tx, payloadSize, n int) error {
+		if err := insertMany(ctx, tx, "key", payloadSize, n); err != nil {
+			return err
+		}
+		b.Log("insert", n)
```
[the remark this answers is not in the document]
marco6: Removed.

test/list_test.go
marco6: The setup is not making any channel. It's setting up the scenario by bulk-inserting randomly generated rows, so that we can benchmark over a big dataset. 

Doing that with the client would make a test like this take dramatically longer (event 10-50x) and wouldn't be very useful.

test/list_test.go
marco6: Yeah we don't really care about the key, but we just want to exercise the database on different ranges (to avoid caching effects), so that we know the "random access" speed of the query.

Regarding the `\x01`, I agree it is weird, but that weirdness *comes from etcd* (see for example here for how [kubernetes uses pagination](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L326)). You will notice though that we are using `\x01` instead of `\x00`. The reason for that is a [bug in dqlite protocol](https://dqlite.io/docs/reference/wire-protocol) which is hard to fix without breaking backward compatibility. The dqlite team is already aware of that.

test/admission_test.go
```diff
@@ -42,7 +47,7 @@ func TestAdmissionControl(t *testing.T) {
 			var numSuccessfulWriterTxn = atomic.Uint64{}
 			var numSuccessfulReaderTxn = atomic.Uint64{}
 
-			reader := func(first int, last int) {
+			read := func(first int, last int) {
```
TheSignPainter98, line 50: On re-reading, it's not quite clear what `first` and `last` are. Could these be made more descriptive, like `first_key_num` and `last_key_num` for example? (Not sure about these specific names)

Also, as these are both `int`s, the parameters can share the same type declaration (not sure how to describe it, I mean `first, last int`)

test/admission_test.go
```diff
@@ -77,35 +82,30 @@ func TestAdmissionControl(t *testing.T) {
...
-			writers := 500
-			writers_replication := 10
-			write_entries := 1000 / writers
-			wg.Add(readers*readers_replication + writers*writers_replication)
+			readers := writeLimit * 4
+			read_entries := 20
+
+			writers := writeLimit * 4
+			write_entries := 20
+			wg.Add(readers + writers)
```
TheSignPainter98, line 90: This `wg.Add` doesn't feel more or less strongly associated with the readers or the writers, so I think there should be a newline here, there shouldn't be one between `read_entries` and `writers`

test/list_test.go
```diff
@@ -178,32 +178,95 @@ func TestList(t *testing.T) {
...
+			name string
+			size int
+		}{{
+			name: "tiny",
+			size: 100,
+		}, {
+			name: "page-fitting",
+			size: 1000,
+		}, {
+			name: "page-overflowing",
```
TheSignPainter98, line 206: Somehow these read like adjectives, but I think these should instead be `fits-in-page` and `overflows-page`

test/list_test.go
TheSignPainter98: Ah this is unfortunate, thanks for the info!

test/list_test.go
```diff
@@ -178,32 +178,95 @@ func TestList(t *testing.T) {
...
-			b.StopTimer()
-			g := NewWithT(b)
+	setup := func(ctx context.Context, tx *sql.Tx, payloadSize, n int) error {
+		if err := insertMany(ctx, tx, "key", payloadSize, n); err != nil {
+			return err
+		}
+		if err := updateMany(ctx, tx, "key", payloadSize, n/2); err != nil {
+			return err
+		}
+		if err := deleteMany(ctx, tx, "key", n/2); err != nil {
```
TheSignPainter98, line 189: If I understand correctly, this will result in a database of size `n/2`, should these parameters with `n` be doubled?

TheSignPainter98 (changes requested):
Looking good, just a couple of trivials then this looks ready to me

TheSignPainter98 (approved):
Very nice, thanks for the changes!

PR #140, opened 2024-07-22
Improve list query

marco6:
This PR changes slightly the index to make it covering when querying for latest rows.

It also changes the List query (and all those related to the same index) so that it is possible to exploit better the properties of this updated index. While doing that, it removes a redundant query.


marco6:
Depends on #136, #132 and #131 .

marco6:
This PR also depends on #141

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -73,13 +73,15 @@ type Dialect interface {
 	GetSize(ctx context.Context) (int64, error)
 	GetCompactInterval() time.Duration
 	GetPollInterval() time.Duration
-	Close()
+	Close() error
```
louiseschmidtgen, line 76: Nice, thanks for adding the error check!

marco6:
During the development of #170 we found out that List queries cannot return PrevKV. Moreover, Update does not use `get` to retrieve the event anymore.

Delete query is on its way to be updated as well. This means that all GET path do not really need the prevKV and we could improve the read path way more by not reading the previous value (less data to read, less allocations).

There is still a place where this might happen: the watch query. Watchers might ask for prevKV. In this case we need to understand how to deal with that request (if necessary).

marco6:
Depends on #195 as it removes PrevKV from the query.

marco6:
Depends on #180.

marco6:
Dropping dependency from #180. This is not ready to be merged yet though as we don't have a path forward for the migration yet.

marco6:
Superseded by #281

PR #141, opened 2024-07-22
Enable minor schema version updates

louiseschmidtgen:
### Description
Enable minor schema version updates.

### DIFF
--- a/pkg/kine/drivers/sqlite/schema.go
+++ b/pkg/kine/drivers/sqlite/schema.go
@@ -3,6 +3,8 @@ package sqlite
 import (
 	"context"
 	"database/sql"
+
+	"github.com/pkg/errors"
 )
 
 // The database version that designates whether table migration
@@ -13,9 +15,46 @@ import (
 // already
 
 const (
-	databaseSchemaMajorVersion = 1
-	databaseSchemaMinorVersion = 0
+	databaseSchemaMajorVersion int16 = 1
+	databaseSchemaMinorVersion int16 = 0
+	mask                       int32 = 0xFFFF // Mask for 2 bytes (16 bits)
 )
+
+type schemaVersion int32
+
+var (
+	databaseSchemaVersion = ToSchemaVersion(databaseSchemaMajorVersion, databaseSchemaMinorVersion)
+)
+
+func ToSchemaVersion(major int16, minor int16) schemaVersion {
+	return schemaVersion(int32(major)<<16 | int32(minor))
+}
+
+func (sv schemaVersion) Major() int16 {
+	// Extract the high 16 bits
+	return int16((int32(sv) >> 16) & mask)
+}
+
+func (sv schemaVersion) Minor() int16 {
+	// Extract the lower 16 bits
+	return int16(int32(sv) & mask)
+}
+
+func (sv schemaVersion) CanMigrate(targetSV schemaVersion) (bool, error) {
+	// Check wether version is the same
+	if sv.Major() == targetSV.Major() && sv.Minor() == targetSV.Minor() {
+		return false, nil
+	}
+	// Major version must be the same
+	if sv.Major() != targetSV.Major() {
+		return false, errors.Errorf("can not migrate between different major versions: %d to %d", sv.Major(), targetSV.Major())
+	}
+	// Minor version must be greater
+	if sv.Minor() > targetSV.Minor() {
+		return false, errors.Errorf("can not rollback to earlier minor version: %d to %d", sv.Minor(), targetSV.Minor())
+	}
+	return true, nil
+}
 
 // applySchemaV1 moves the schema from version 0 to version 1,
 // taking into account the possible unversioned schema from
--- a/pkg/kine/drivers/sqlite/schema.go
+++ b/pkg/kine/drivers/sqlite/schema.go
@@ -13,52 +13,42 @@ import (
 // the latest revisions must have been recorded in the Kine table
 // already
 
-const (
-	databaseSchemaMajorVersion int16 = 0
-	databaseSchemaMinorVersion int16 = 1
-	mask                       int32 = 0xFFFF // Mask for 2 bytes (16 bits)
-)
-
 type SchemaVersion int32
 
 var (
-	databaseSchemaVersion = ToSchemaVersion(databaseSchemaMajorVersion, databaseSchemaMinorVersion)
+	databaseSchemaVersion = NewSchemaVersion(1, 0)
 )
 
-func ToSchemaVersion(major int16, minor int16) SchemaVersion {
+func NewSchemaVersion(major int16, minor int16) SchemaVersion {
 	return SchemaVersion(int32(major)<<16 | int32(minor))
 }
 
 func (sv SchemaVersion) Major() int16 {
 	// Extract the high 16 bits
-	return int16((int32(sv) >> 16) & mask)
+	return int16((int32(sv) >> 16))
 }
 
 func (sv SchemaVersion) Minor() int16 {
 	// Extract the lower 16 bits
-	return int16(int32(sv) & mask)
+	return int16(sv)
 }
 
-func (sv SchemaVersion) CanMigrate(targetSV SchemaVersion) (bool, error) {
-	// Check wether version is the same
-	if sv.Major() == targetSV.Major() && sv.Minor() == targetSV.Minor() {
-		return false, nil
-	}
+func (sv SchemaVersion) CompatibleWith(targetSV SchemaVersion) error {
 	// Major version must be the same
 	if sv.Major() != targetSV.Major() {
-		return false, fmt.Errorf("can not migrate between different major versions")
+		return fmt.Errorf("can not migrate between different major versions")
 	}
 	// Minor version must be greater
 	if sv.Minor() > targetSV.Minor() {
-		return false, fmt.Errorf("can not rollback to earlier minor version")
+		return fmt.Errorf("can not rollback to earlier minor version")
 	}
-	return true, nil
+	return nil
 }
 
-// applySchemaV1 moves the schema from version 0 to version 1,
+// applySchemaV0_1 moves the schema from version 0 to version 1,
 // taking into account the possible unversioned schema from
 // upstream kine.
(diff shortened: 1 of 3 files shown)


pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -148,34 +153,43 @@ func setup(ctx context.Context, db *sql.DB) error {
...
-	setUserVersionSQL := fmt.Sprintf(`PRAGMA user_version = %d`, databaseSchemaVersion)
+	setUserVersionSQL := fmt.Sprintf(`PRAGMA user_version = %d`, toInt32(databaseSchemaMajorVersion, databaseSchemaMinorVersion))
 	if _, err := txn.ExecContext(ctx, setUserVersionSQL); err != nil {
 		return err
 	}
 
 	return nil
 }
 
+func getMajorMinorVersion(value int32) (major int16, minor int16) {
```
marco6, line 178: Can we have a type for the version? Something like:

```go
type schemaVersion int

func (sv schemaVersion) Major() int
func (sv schemaVersion) Minor() int

func MakeSchemaVersion(major, minor int) schemaVersion
```

You could also add a method like:

```go
func (sv schemaVersion)  CompatibleWith(other schemaVersion) error
```

Which checks if two schema version are compatible.

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -121,15 +121,20 @@ func NewVariant(ctx context.Context, driverName, dataSourceName string) (server.
...
 	row := db.QueryRowContext(ctx, `PRAGMA user_version`)
 	if err := row.Scan(&schemaVersion); err != nil {
 		return err
 	}
+	currentMajor, currentMinor := getMajorMinorVersion(schemaVersion)
 
-	if schemaVersion == databaseSchemaVersion {
+	if currentMajor == databaseSchemaMajorVersion && currentMinor == databaseSchemaMinorVersion {
 		return nil
+	} else if currentMajor > 0 && currentMajor < databaseSchemaMajorVersion {
```
marco6, line 134: why `currentMajor` should always be above 0?

marco6 (changes requested):
Thanks for these changes Louise! Below few comments:

pkg/kine/drivers/sqlite/schema.go
```diff
@@ -11,7 +11,11 @@ import (
 // present anymore, and unexpired rows of the key_value table with
 // the latest revisions must have been recorded in the Kine table
 // already
-const databaseSchemaVersion = 1
+
+const (
+	databaseSchemaMajorVersion = 1
```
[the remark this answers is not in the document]
louiseschmidtgen: Ah, I thought we were on major 1 already. I've changed it.

pkg/kine/drivers/sqlite/schema.go
```diff
@@ -13,10 +14,47 @@ import (
...
+	mask                       int32 = 0xFFFF // Mask for 2 bytes (16 bits)
 )
 
+type SchemaVersion int32
+
+var (
+	databaseSchemaVersion = ToSchemaVersion(databaseSchemaMajorVersion, databaseSchemaMinorVersion)
+)
+
+func ToSchemaVersion(major int16, minor int16) SchemaVersion {
```
marco6, line 28: Can we call this `NewSchemaVersion` instead? The `ToXXX` doesn't have a subject. In my mind it would make sense if you already had an object so that you call `yyy.ToXXX` but in this case, you are creating a new one.

pkg/kine/drivers/sqlite/schema.go
```diff
@@ -13,10 +14,47 @@ import (
 // already
 
 const (
-	databaseSchemaMajorVersion = 1
-	databaseSchemaMinorVersion = 0
+	databaseSchemaMajorVersion int16 = 0
```
marco6, line 17: I think we can safely inline these 2 constants in the definition of `databaseSchemaVersion` below, so that someone reading that line immediately knows the version.

pkg/kine/drivers/sqlite/schema.go
```diff
@@ -13,10 +14,47 @@ import (
...
+	databaseSchemaVersion = ToSchemaVersion(databaseSchemaMajorVersion, databaseSchemaMinorVersion)
+)
+
+func ToSchemaVersion(major int16, minor int16) SchemaVersion {
+	return SchemaVersion(int32(major)<<16 | int32(minor))
+}
+
+func (sv SchemaVersion) Major() int16 {
+	// Extract the high 16 bits
+	return int16((int32(sv) >> 16) & mask)
```
marco6, line 34: The masking is not needed as the conversion to int16 will remove upper bits.

pkg/kine/drivers/sqlite/schema.go
```diff
@@ -13,10 +14,47 @@ import (
 // already
 
 const (
-	databaseSchemaMajorVersion = 1
-	databaseSchemaMinorVersion = 0
+	databaseSchemaMajorVersion int16 = 0
+	databaseSchemaMinorVersion int16 = 1
+	mask                       int32 = 0xFFFF // Mask for 2 bytes (16 bits)
```
marco6, line 19: This package-wide variable should be removed or renamed with a more specific name.

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -121,20 +121,15 @@ func NewVariant(ctx context.Context, driverName, dataSourceName string) (server.
...
 	}
-	currentMajor, currentMinor := getMajorMinorVersion(schemaVersion)
 
-	if currentMajor == databaseSchemaMajorVersion && currentMinor == databaseSchemaMinorVersion {
-		return nil
-	} else if currentMajor > 0 && currentMajor < databaseSchemaMajorVersion {
-		return errors.Errorf("can not update major versions: %d to %d", currentMajor, databaseSchemaMajorVersion)
-	} else if currentMajor > databaseSchemaMajorVersion {
-		return errors.Errorf("can not rollback to earlier major version: %d to %d", currentMajor, databaseSchemaMajorVersion)
+	if canMigrate, err := currentSchemaVersion.CanMigrate(databaseSchemaVersion); !canMigrate {
```
marco6, line 131: I'd prefer to have a more explicit logic here. Consider that in this place 2 things are decided:
 - if we should run migration;
 - if we can work with the current schema (for example, the current schema is a newer version, but it is compatible). 
 
 So, I would prefer something on the lines of:
 
 ```go
 if err := databaseSchemaVersion.CompatibleWith(currentSchemaVersion); err != nil {
   return err
 }
 if currentSchemaVersion >= databaseSchemaVersion {
   // Already to the latest version.
   return nil
 }
 ```
 
 This would change the `CanMigrate` method on the `SchemaVersion` which is very specific to our migration logic (which might change) to a much more generic `CompatibleWith` which is a generic method over semantic versioning.

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -153,43 +148,33 @@ func setup(ctx context.Context, db *sql.DB) error {
...
 func migrate(ctx context.Context, txn *sql.Tx) error {
-	var userVersion int32
+	var currentVersion SchemaVersion
 
 	row := txn.QueryRowContext(ctx, `PRAGMA user_version`)
-	if err := row.Scan(&userVersion); err != nil {
+	if err := row.Scan(&currentVersion); err != nil {
 		return err
 	}
-	currentMajor, currentMinor := getMajorMinorVersion(int32(userVersion))
```
marco6, line 157: We need to add a check for version compatibility here as well as someone might have upgraded in the meanwhile.

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -153,43 +148,33 @@ func setup(ctx context.Context, db *sql.DB) error {
...
 
-	if currentMajor == 0 && currentMinor == 0 {
-		if err := applySchemaV1(ctx, txn); err != nil {
-			return err
+	switch currentVersion.Major() {
```
marco6, line 158: This should switch over the entire version, not just the minor. You can use in the switch case the version.

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -153,43 +148,33 @@ func setup(ctx context.Context, db *sql.DB) error {
...
+	case 0:
+		for v := currentVersion.Minor(); v <= databaseSchemaMinorVersion; v++ {
```
marco6, line 160: This loop is not necessary.

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -153,43 +148,33 @@ func setup(ctx context.Context, db *sql.DB) error {
...
+			switch v {
+			case 0:
+				if err := applySchemaV1(ctx, txn); err != nil {
```
marco6, line 163: Would it be possible to rename the `applySchemaXXX` methods to follow semver? Something like `applySchemaV0_1`

marco6 (changes requested):
Thanks for the changes! One more pass:

pkg/kine/drivers/sqlite/schema.go
louiseschmidtgen: Just `databaseSchemaVersion = NewSchemaVersion(1, 0)`?

pkg/kine/drivers/sqlite/schema.go
louiseschmidtgen: it's removed :)

pkg/kine/drivers/sqlite/sqlite.go
louiseschmidtgen: Yes, that seems like a better way!

pkg/kine/drivers/sqlite/schema.go
marco6: Yes

louiseschmidtgen: I realized it's in the last commit already :D

marco6 (approved):
LGTM

PR #143, opened 2024-07-25
Sqllog otel traces

louiseschmidtgen:
### Description

This PR adds further otel traces to sqllog functions.

### DIFF
--- a/pkg/kine/logstructured/sqllog/sql.go
+++ b/pkg/kine/logstructured/sqllog/sql.go
@@ -568,6 +568,11 @@ func (s *SQLLog) Count(ctx context.Context, prefix, startKey string, revision in
 		span.RecordError(err)
 		span.End()
 	}()
+	span.SetAttributes(
+		attribute.String("prefix", prefix),
+		attribute.String("startKey", startKey),
+		attribute.Int64("revision", revision),
+	)
 	if revision == 0 {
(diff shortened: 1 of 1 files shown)


pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -539,6 +562,12 @@ func canSkipRevision(rev, skip int64, skipTime time.Time) bool {
...
 
 func (s *SQLLog) Count(ctx context.Context, prefix, startKey string, revision int64) (int64, int64, error) {
+	var err error
+	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Count", otelName))
+	defer func() {
+		span.RecordError(err)
+		span.End()
+	}()
 	if revision == 0 {
 		return s.d.CountCurrent(ctx, prefix, startKey)
```
addyess, line 572: Is it not important in this method to add any attributes?
```suggestion
	}()
	span.SetAttributes(
		attribute.String("prefix", prefix),
		attribute.String("startKey", startKey),
		attribute.Int64("revision", revision),
    )
	if revision == 0 {
		return s.d.CountCurrent(ctx, prefix, startKey)
```

addyess (commented):
I found one place where it seems attributes were missing.  Was that intentional?

addyess:
thanks @marco6 if you can agree!

Issue #146, opened 2024-07-26
Otel trace: Invalid compaction range & non-5min compaction interval


louiseschmidtgen:
## Description
On average the idle cluster compaction loop runs for 6s.
Invalid compaction range & non-5min compaction interval after rogue 7m compaction run

## Version 
```
moonray: v1.30.3 rev 729, chaos deployed
go: go1.22.5
sqlite: 3.40.0 (compiled), 3.40.0 (linked)
dqlite: 1.14.0 (compiled), 1.14.0 (linked)
libuv: 1.48.0 (compiled), 1.48.0 (linked)
```
## Recorded traces
![Screenshot from 2024-07-26 14-28-49](https://github.com/user-attachments/assets/fcc0293c-8a31-4abe-b4e9-07f3d1f64756)
![Screenshot from 2024-07-26 14-30-12](https://github.com/user-attachments/assets/de4dff3c-8ba6-41c4-a43b-45f513b4ae15)

louiseschmidtgen:
Addressed by https://github.com/canonical/k8s-dqlite/pull/149

Issue #145, opened 2024-07-26
Otel trace: compaction run (7m 21s)


louiseschmidtgen:
## Description
I caught a compaction run taking 7m 21s. 
## Version 
```
moonray: v1.30.3 rev 729, chaos deployed
go: go1.22.5
sqlite: 3.40.0 (compiled), 3.40.0 (linked)
dqlite: 1.14.0 (compiled), 1.14.0 (linked)
libuv: 1.48.0 (compiled), 1.48.0 (linked)
```
## Recorded Trace
![Screenshot from 2024-07-26 14-21-12](https://github.com/user-attachments/assets/1ba075e2-a360-4692-a690-95c7be3c82bb)
![Screenshot from 2024-07-26 14-30-12](https://github.com/user-attachments/assets/9a95f6fe-5ba6-4acb-bf92-a868078ce533)
![Screenshot from 2024-07-26 14-30-12](https://github.com/user-attachments/assets/acf00d5e-135d-4f99-9700-b3e80570c87b)

## Possibly related:
These traces have been recorded around the time of this rogue compaction:
Invalid compaction range & non-5min compaction interval: https://github.com/canonical/k8s-dqlite/issues/146
logstructured.delete 10.77s https://github.com/canonical/k8s-dqlite/issues/147

Note: We are not seeing any traces from current revision/ get compact revision instead the first span is delete_revision.

louiseschmidtgen:
Addressed by https://github.com/canonical/k8s-dqlite/pull/149

PR #149, opened 2024-07-29
Improve compaction

marco6:
## Description
This PR rewrites the compaction step in a similar way [upstream](https://github.com/k3s-io/kine/pull/51) is doing. This way, the throughput should be a lot higher. The main differences from upstream are:
 - small batches instead of a single big one;
 - proper range boundaries to reduce search space;
 - in-transaction update of the compaction version.

This query does not use indexes right now. It's still *way faster* than what we have and what upstream has. Maybe we could improve things with a covering index if/when it's needed for the Watch/TTL query. Otherwise we might be ok with what we have.

## Improvements to the previous implementation
Up to this point we've been doing compaction on each revision within the compaction range row by row. Looping over all revisions takes about 7s on an idle cluster, rare issues have been recorded where compaction takes minutes. With this compaction query improvement we are doing a compaction batch in around 40ms and a max <100ms on an idle cluster. (Note: the idle cluster only needs 1 batch to complete the compaction under normal circumstances).

This PR addresses: https://github.com/canonical/k8s-dqlite/issues/145 and https://github.com/canonical/k8s-dqlite/issues/146

### DIFF
--- a/pkg/kine/logstructured/sqllog/sql.go
+++ b/pkg/kine/logstructured/sqllog/sql.go
@@ -145,18 +145,6 @@ func (s *SQLLog) DoCompact(ctx context.Context) error {
 		return fmt.Errorf("failed to initialise compaction: %v", err)
 	}
 
-	// NOTE: Upstream is ignoring the last 1000 revisions, however that causes the following CNCF conformance test to fail.
-	// This is because of low activity, where the created list is part of the last 1000 revisions and is not compacted.
-	// Link to failing test: https://github.com/kubernetes/kubernetes/blob/f2cfbf44b1fb482671aedbfff820ae2af256a389/test/e2e/apimachinery/chunking.go#L144
-	// To address this, we only ignore the last 100 revisions instead
-	rev, err := s.d.CurrentRevision(ctx)
-	if err != nil {
-		return err
-	}
-	return s.d.Compact(ctx, rev-SupersededCount)
-}
-
-func (s *SQLLog) compact(ctx context.Context) error {
 	// When executing compaction as a background operation
 	// it's best not to take too much time away from query
 	// operation and similar. As such, we do compaction in
@@ -167,8 +155,16 @@ func (s *SQLLog) compact(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	// NOTE: Upstream is ignoring the last 1000 revisions, however that causes the following CNCF conformance test to fail.
+	// This is because of low activity, where the created list is part of the last 1000 revisions and is not compacted.
+	// Link to failing test: https://github.com/kubernetes/kubernetes/blob/f2cfbf44b1fb482671aedbfff820ae2af256a389/test/e2e/apimachinery/chunking.go#L144
+	// To address this, we only ignore the last 100 revisions instead
+	target -= SupersededCount
 	for start < target {
 		batchRevision := start + compactBatchSize
+		if batchRevision > target {
+			batchRevision = target
+		}
 		if err := s.d.Compact(s.ctx, batchRevision); err != nil {
 			return err
 		}
@@ -354,7 +350,7 @@ func (s *SQLLog) startWatch() (chan interface{}, error) {
 			case <-s.ctx.Done():
 				return
 			case <-t.C:
-				if err := s.compact(s.ctx); err != nil {
+				if err := s.DoCompact(s.ctx); err != nil {
 					logrus.WithError(err).Trace("compaction failed")
 				}
 			}
--- a/pkg/kine/logstructured/sqllog/sql.go
+++ b/pkg/kine/logstructured/sqllog/sql.go
@@ -140,7 +140,12 @@ func (s *SQLLog) compactStart(ctx context.Context) error {
 
 // DoCompact makes a single compaction run when called. It is intended to be called
 // from test functions that have access to the backend.
-func (s *SQLLog) DoCompact(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


pkg/kine/drivers/sqlite/schema.go
```diff
@@ -61,6 +61,19 @@ CREATE TABLE kine
 	return nil
 }
 
+// applySchemaV2 moves the schema from version 1 to version 2
+func applySchemaV2(ctx context.Context, txn *sql.Tx) error {
```
louiseschmidtgen, line 65: Let's make it applySchemaV0_2 and please rebase of master with my new schema version updates.

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -12,14 +12,14 @@ import (
...
-	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 )
 
 const (
-	SupersededCount = 100
-	otelName        = "sqllog"
+	SupersededCount  = 100
+	compactBatchSize = 1000
```
louiseschmidtgen, line 21: Can we add a compaction txn timeout?

marco6:
Depends on #140

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,233 @@
+package prepared
```
louiseschmidtgen, line 1: Should we move this file under generic?

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,233 @@
...
+	// lruList is implemented as a circular linked list
+	head *lruEntry
+}
+
+type lruEntry struct {
+	query string
+	stmt  *stmt
+
+	// Adjacent entries, unless at the tail of the list, the
+	// next entry was added less recently tham the current.
```
louiseschmidtgen, line 197:
```suggestion
	// next entry was added less recently than the current.
```

pkg/kine/prepared/db.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	defer stmt.Close()
+
+	return tx.tx.StmtContext(ctx, stmt.Stmt).QueryContext(ctx, args...)
+}
+
+func (tx *Tx) Commit() error   { return tx.tx.Commit() }
+func (tx *Tx) Rollback() error { return tx.tx.Rollback() }
+
+type lruList struct {
```
louiseschmidtgen, line 187: I like moving the query preparation to this package! Could you please add a description here about the purpose of the lruList and what lru stands for?

test/admission_test.go
```diff
@@ -29,11 +31,14 @@ func TestAdmissionControl(t *testing.T) {
 				backendType: backendType,
 				endpointParameters: []string{
 					"admission-control-policy=limit",
-					"admission-control-policy-limit-max-concurrent-txn=600",
+					fmt.Sprintf("admission-control-policy-limit-max-concurrent-txn=%d", writeLimit),
```
louiseschmidtgen, line 34: Any particular scenario causing you to drop this limit?

louiseschmidtgen (commented):
Great work Marco, excited to watch this land! Let's test this against the k8s-snap and microk8s and view traces as discussed.

pkg/kine/prepared/db.go
marco6: I want this to not be part of a driver so that if we have a new implementation (and we will, hopefully), then we can share this package.

pkg/kine/prepared/db.go
marco6: All the lru has been removed as probably not very useful and yet quite complex.

pkg/kine/prepared/db.go
marco6: Removed.

test/admission_test.go
marco6: This might actually be a good candidate for a separate PR.

The problem is that this test creates hundreds of goroutines just to make sure that admission control blocks some queries. However the same check can be achieved with a smaller number without loosing correctness.

This change brings the execution of the test from minutes (on my machine) to less then a second. And in general, it is important to be able to run tests frequently while developing.

marco6:
Marking as ready since #136 was meged.

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -18,8 +18,9 @@ import (
...
```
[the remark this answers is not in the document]
louiseschmidtgen: That's a great point @bschimke95 
IMHO we should be fine with this default batch size. Note that kine does one big transaction here instead of batches which can block a db. These batches address those high workload scenarios and 1000 revisions should still be comfortably manageable under high load. 

Nice-to-know:
I can't find anywhere upstream etcd is setting a compaction batch size, but they also have some sort of logic in place where they can stop doing compaction under high load. I think this comes from the etcdserver request and we are not compacting on this request but simply every 5 minutes or whatever our compaction interval is. If I remember correctly, the compaction interval can be tweaked (the time between each compaction). 
This could be worth investigating as part of future work.

Overall, if we do run into issues under high load, I think we will need a change that goes beyond changing the batch size.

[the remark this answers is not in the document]
bschimke95: cool, thanks for the context

bschimke95 (approved):

PR #152, opened 2024-08-01
Improve create query

louiseschmidtgen:
### Description
This PR improves the create method by moving the logic into one query.

### Query
The create query:
- If there is no previous entry for a given key the query inserts the new row with the key and its value.
- If there is a previous entry for a given key exists:
  -  where the previous entry is not a deleted entry, the query inserts the new row with the key and its value.
  -  where the previous entry is a deleted entry, the query does nothing.

Then we check if any rows have been affected by the query, if not we return a key exists error.

### Index
Note: we will make an update to the `kine_name_index` when we update the schema version to include `delete`. This will allow us to have a covering index for the new create query.

### DIFF
--- a/pkg/kine/logstructured/sqllog/sql.go
+++ b/pkg/kine/logstructured/sqllog/sql.go
@@ -620,20 +620,19 @@ func (s *SQLLog) Append(ctx context.Context, event *server.Event) (int64, error)
 	return rev, nil
 }
 
-func (s *SQLLog) Create(ctx context.Context, key string, value []byte, lease int64) (int64, error) {
-	var err error
+func (s *SQLLog) Create(ctx context.Context, key string, value []byte, lease int64) (rev int64, err error) {
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Create", otelName))
 	defer func() {
 		span.RecordError(err)
+		span.SetAttributes(attribute.Int64("revision", rev))
 		span.End()
 	}()
 	span.SetAttributes(attribute.String("key", key))
 
-	rev, err := s.d.Create(ctx, key, value, lease)
+	rev, err = s.d.Create(ctx, key, value, lease)
 	if err != nil {
 		return 0, err
 	}
-	span.SetAttributes(attribute.Int64("revision", rev))
 
 	select {
 	case s.notify <- rev:
--- a/pkg/kine/logstructured/sqllog/sql.go
+++ b/pkg/kine/logstructured/sqllog/sql.go
@@ -64,6 +64,7 @@ type Dialect interface {
 	AfterPrefix(ctx context.Context, prefix string, rev, limit int64) (*sql.Rows, error)
 	After(ctx context.Context, rev, limit int64) (*sql.Rows, error)
 	Insert(ctx context.Context, key string, create, delete bool, createRevision, previousRevision int64, ttl int64, value, prevValue []byte) (int64, error)
+	Create(ctx context.Context, key string, value []byte, lease int64) (int64, error)
 	GetRevision(ctx context.Context, revision int64) (*sql.Rows, error)
 	DeleteRevision(ctx context.Context, revision int64) error
 	GetCompactRevision(ctx context.Context) (int64, int64, error)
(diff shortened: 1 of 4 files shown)


louiseschmidtgen:
## Benchmark Analysis
The create query takes less (because of the network most likely):
```
                    │    Before    │                After                │
                    │    sec/op    │    sec/op     vs base               │
Create/sqlite-4       1.548m ± 17%   1.443m ± 15%        ~ (p=0.053 n=7)
Create/dqlite-4       1.427m ± 13%   1.272m ± 14%  -10.88% (p=0.026 n=7)
```
The total amount of pages read for the transaction decreased by quite a lot:
```
                    │    Before     │                After                 │
                    │ page-reads/op │ page-reads/op  vs base               │
Create/sqlite-4         21.27 ±  2%    13.92 ±   4%  -34.56% (p=0.001 n=7)
Create/dqlite-4         16.75 ±  3%    10.10 ±   2%  -39.70% (p=0.001 n=7)
```
The transaction is spending way less time reading than before. Consider that this most likely compensates the increase in time spent writing (it spent additional 16µs on the write as you pointed out, but 35µs less during the reads).
```
                    │     Before      │                  After                  │
                    │ sec-reading/op  │ sec-reading/op  vs base                 │
Create/sqlite-4        91.30µ ± 23%      43.70µ ± 879%  -52.14% (p=0.026 n=7)
Create/dqlite-4       44.700µ ± 19%      9.100µ ±  55%  -79.64% (p=0.001 n=7)
```
Allocates way less both in number of allocations and in size:
```
                    │     Before     │                After                 │
                    │      B/op      │     B/op       vs base               │
Create/sqlite-4        23.81Ki ±  0%   19.31Ki ±  2%  -18.90% (p=0.001 n=7)
Create/dqlite-4        21.55Ki ±  0%   17.90Ki ±  0%  -16.91% (p=0.001 n=7)
                    │  allocs/op   │  allocs/op    vs base                 │
Create/sqlite-4        535.0 ±  0%    424.0 ±  4%  -20.75% (p=0.001 n=7)
Create/dqlite-4        397.0 ±  0%    336.0 ±  0%  -15.37% (p=0.001 n=7)
```
We write and read dramatically less from/to the network:
```
                    │        Before         │                    After                     │
                    │ network-bytes-read/op │ network-bytes-read/op  vs base               │
Create/dqlite-4                 423.3 ±  0%             239.4 ±  0%  -43.44% (p=0.001 n=7)
                    │          Before          │                      After                      │
                    │ network-bytes-written/op │ network-bytes-written/op  vs base               │
Create/dqlite-4                  1040.0 ±   0%               313.0 ±   0%  -69.90% (p=0.001 n=7)
```

louiseschmidtgen:
## Traces
Traces confirm that our k8s cluster is happy and our create query works as expected:
Screenshot from 2024-08-01 19-09-48

bschimke95 (commented):
code-wise LGMT
Not sure about the query improvements, I let @marco6 give the final approval on this.

I see that this decreases read time and amount of reads but wouldn't jump straight to the conclusion that the decrease in reads outweighs the increase in write time.
This highly depends on the workload.
As said, up to @marco6 to decide.

marco6:
> code-wise LGMT Not sure about the query improvements, I let @marco6 give the final approval on this.
> 
> I see that this decreases read time and amount of reads but wouldn't jump straight to the conclusion that the decrease in reads outweighs the increase in write time. This highly depends on the workload. As said, up to @marco6 to decide.

First of all, I want to point out that the time that increased is the "time spent in a write transaction", which is slightly different from the "time spent writing".

To me, it is a clear improvement. If we look at the performance in Dqlite we get a nice 10% decrease of latency. All resource usage is decreasing sharply. I would accept that even if the latency would increase TBH.

As a side note, this query is "incomplete" in a sense: the query can be made dramatically faster with a slight change in one of the indexes. Most likely, in that case, we could even remove this increase in time spent in a write transaction.

So, green light from my part, but maybe @ktsakalozos has a different opinion? I.e. we might want to decide that it's better to wait for an update in the index?

As a way to set up our minds, we could have a run with kube burner like we were doing in the last iteration so that we understand better and then decide.

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -619,6 +620,27 @@ func (s *SQLLog) Append(ctx context.Context, event *server.Event) (int64, error)
...
+	}()
+	span.SetAttributes(attribute.String("key", key))
+
+	rev, err = s.d.Create(ctx, key, value, lease)
+	if err != nil {
+		return 0, err
+	}
+
+	select {
+	case s.notify <- rev:
```
louiseschmidtgen, line 638: This notifies the watcher on the creation of this revision. This is part of the original implementation, we can change it in a future update of the watcher.

PR #153, opened 2024-08-07
Build raft with dqlite

louiseschmidtgen:
## Description 
The raft repo is archived we should be building Raft with dqlite.


neoaggelos (approved):
LGTM, after a green CI

Issue #138, opened 2024-07-22
Close metrics connection


louiseschmidtgen:
We do not shut down the metric server: https://github.com/canonical/k8s-dqlite/blob/master/cmd/root.go#L80

Issue #137, opened 2024-07-22
Disable metrics endpoint by default


louiseschmidtgen:
K8s-dqlite collects metrics by default. 

We should not have these metrics collected by default in production settings:
`Enable metrics endpoint                       address="127.0.0.1:9042"`

Note: This is part of the upstream metrics collection and not part of our otel observability effort.

PR #154, opened 2024-08-07
Disable kine metrics by default

louiseschmidtgen:
## Description
Disable kine metrics by default and shutdown the server. 

Closes https://github.com/canonical/k8s-dqlite/issues/138 https://github.com/canonical/k8s-dqlite/issues/137

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -143,7 +143,7 @@ var (
 					logrus.WithError(err).Warning("Failed to shutdown OpenTelemetry SDK")
 				}
 			}
-			if rootCmdOpts.metrics && metricsServer != nil {
+			if metricsServer != nil {
 				if err := metricsServer.Shutdown(stopCtx); err != nil {
 					logrus.WithError(err).Fatal("Failed to shutdown metrics endpoint")
 				}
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -130,6 +143,11 @@ var (
 					logrus.WithError(err).Warning("Failed to shutdown OpenTelemetry SDK")
 				}
 			}
+			if rootCmdOpts.metrics && metricsServer != nil {
```
bschimke95, line 146: I guess you don't need to check for the cmd opts because `metricsServer` would be `nil` if the flag is `false`.

cmd/root.go
```diff
@@ -152,15 +170,15 @@ func init() {
...
 	rootCmd.Flags().StringVar(&rootCmdOpts.lowAvailableStorageAction, "low-available-storage-action", "none", "Action to perform in case the available storage is low. One of (none|handover|terminate). none means no action is performed. handover means the dqlite node will handover its leadership role, if any. terminate means this dqlite node will shutdown")
 	rootCmd.Flags().StringVar(&rootCmdOpts.admissionControlPolicy, "admission-control-policy", "allow-all", "Transaction admission control policy to use. One of (allow-all|limit-concurrent-transactions). Set to allow-all to disable the admission control")
-	// TODO(MK-1408): This value is highly dependend on underlying hardware, thus making the default value a bit useless. The linked card will implement a dynamic way to set this value.
+	// TODO(MK-1408): This value is highly depends on underlying hardware, thus making the default value a bit useless. The linked card will implement a dynamic way to set this value.
```
bschimke95, line 181: I'm not an native speaker but this fix does not seem right :D 
`This value highly depends ...` sound more correct.

louiseschmidtgen: Don't mind me :D

PR #156, opened 2024-08-08
Instrument calls to DB

louiseschmidtgen:
## Description
Add otel traces to the calls to the DB.

### DIFF
--- a/pkg/kine/prepared/db.go
+++ b/pkg/kine/prepared/db.go
@@ -4,8 +4,21 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"fmt"
 	"sync"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
+
+const otelName = "prepared"
+
+var otelTracer trace.Tracer
+
+func init() {
+	otelTracer = otel.Tracer(otelName)
+}
 
 type DB struct {
 	underlying *sql.DB
@@ -22,7 +35,13 @@ func New(db *sql.DB) *DB {
 
 func (db *DB) Underlying() *sql.DB { return db.underlying }
 
-func (db *DB) ExecContext(ctx context.Context, query string, args ...any) (sql.Result, error) {
+func (db *DB) ExecContext(ctx context.Context, query string, args ...any) (result sql.Result, err error) {
+	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.ExecContext", otelName))
+	defer func() {
+		span.RecordError(err)
+		span.End()
+	}()
+
 	stmt, err := db.prepare(ctx, query)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/kine/prepared/db.go
```diff
@@ -22,15 +35,27 @@ func New(db *sql.DB) *DB {
 
 func (db *DB) Underlying() *sql.DB { return db.underlying }
 
-func (db *DB) ExecContext(ctx context.Context, query string, args ...any) (sql.Result, error) {
+func (db *DB) ExecContext(ctx context.Context, query string, args ...any) (result sql.Result, err error) {
```
mateoflorido, line 38: It seems like you are not using your named returns. Perhaps you can modify this to anonymous results again:
```suggestion
func (db *DB) ExecContext(ctx context.Context, query string, args ...any) (sql.Result, error) {
```

pkg/kine/prepared/db.go
```diff
@@ -22,15 +35,27 @@ func New(db *sql.DB) *DB {
...
+
 	stmt, err := db.prepare(ctx, query)
 	if err != nil {
 		return nil, err
 	}
 	return stmt.ExecContext(ctx, args...)
 }
 
-func (db *DB) QueryContext(ctx context.Context, query string, args ...any) (*sql.Rows, error) {
+func (db *DB) QueryContext(ctx context.Context, query string, args ...any) (rows *sql.Rows, err error) {
```
mateoflorido, line 52: Same here:
```suggestion
func (db *DB) QueryContext(ctx context.Context, query string, args ...any) (*sql.Rows, error) {
```

pkg/kine/prepared/db.go
```diff
@@ -22,15 +35,27 @@ func New(db *sql.DB) *DB {
...
+		span.RecordError(err)
+		span.End()
+	}()
+
 	stmt, err := db.prepare(ctx, query)
 	if err != nil {
 		return nil, err
 	}
 	return stmt.ExecContext(ctx, args...)
 }
```
mateoflorido, line 50:
```suggestion
func (db *DB) ExecContext(ctx context.Context, query string, args ...any) (result sql.Result, err error) {
	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.ExecContext", otelName))
	defer func() {
		span.RecordError(err)
		span.End()
	}()

	stmt, err := db.prepare(ctx, query)
	if err != nil {
		return nil, err
	}
	result, err = stmt.ExecContext(ctx, args...)
	return
}
```

pkg/kine/prepared/db.go
louiseschmidtgen: Keeping named returns as they are used in `span.RecordError(err)`

pkg/kine/prepared/db.go
louiseschmidtgen: Keeping named returns as they are used in `span.RecordError(err)`, I would like to avoid using naked returns.

pkg/kine/prepared/db.go
louiseschmidtgen: Keeping named returns as they are used in `span.RecordError(err)`

HomayoonAlimohammadi (approved):
LGTM, Thanks a lot @louiseschmidtgen!

PR #157, opened 2024-08-13
Backport prepared statement into v1.1.9

louiseschmidtgen:
## Description
Back-port prepared statements before query execution avoid using a prepared statement which is prepared on a now closed connection. 
See microk8s issue: https://github.com/canonical/microk8s/issues/4596


louiseschmidtgen:
We should try and keep the patch minimal, keep it at the timeout on the after query.

PR #158, opened 2024-08-14
Make connection pool configurable

louiseschmidtgen:
## Make Connection Pool configurable

We use a connection pool to talk with the database. This connection pool can be configured.
Existing connections are not actively closed by us, we lazily close them via the connection pool configuration.
Through this PR we can set the connection pool configurations via flags.

This PR attempts to mitigate an issue where the leader node is killed and connections are "hanging" on the old connection of the killed node.

## Configuration

Sql allows us to configure these parameters:
- [SetMaxOpenConns](https://pkg.go.dev/database/sql#DB.SetMaxOpenConns)
- [SetMaxIdleConns](https://pkg.go.dev/database/sql#DB.SetMaxIdleConns)
- [SetConnMaxIdleTime](https://pkg.go.dev/database/sql#DB.SetConnMaxIdleTime)
- [SetConnMaxLifetime](https://pkg.go.dev/database/sql#DB.SetConnMaxLifetime)

## Current Configuration

We currently use:
- MaxIdleConns: 5
- MaxOpenConns: 5
- ConnMaxLifetime: 60s
- ConnMaxIdleTime: undefined

## Optimal default values
From a discussion with the dqlite team:
MaxConnections
- lxd team allows only one connection
- benchmarking shows that the best write throughput from dqlite shows you should use at least 2 connections, 5 seems reasonable
- Not certain whether dqlite can take advantage of more than 5 simultaneous connections for reads, if so bumping this number for very large clusters would be advisable.
Max ConnLifetime
- shorter connection lifetimes mean a quicker response to close connections to dead leader nodes, however this can be wasteful on a healthy cluster.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -32,6 +32,11 @@ var (
 		otel                   bool
 		otelAddress            string
 
+		datastoreMaxIdleConnections    int
+		datastoreMaxOpenConnections    int
+		datastoreConnectionMaxLifetime time.Duration
+		datastoreConnectionMaxIdleTime time.Duration
+
 		watchAvailableStorageInterval time.Duration
 		watchAvailableStorageMinBytes uint64
 		lowAvailableStorageAction     string
@@ -108,6 +113,10 @@ var (
 				rootCmdOpts.admissionControlPolicy,
 				rootCmdOpts.acpLimitMaxConcurrentTxn,
 				rootCmdOpts.acpOnlyWriteQueries,
+				rootCmdOpts.datastoreMaxIdleConnections,
+				rootCmdOpts.datastoreMaxOpenConnections,
+				rootCmdOpts.datastoreConnectionMaxLifetime,
+				rootCmdOpts.datastoreConnectionMaxIdleTime,
 				rootCmdOpts.watchQueryTimeout,
 			)
 			if err != nil {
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -32,6 +32,11 @@ var (
 		otel                   bool
 		otelAddress            string
 
+		datastoreMaxIdleConnections    int
```
bschimke95, line 35: maybe `connectionPoolMaxOpenConnections`?

pkg/kine/drivers/generic/generic.go
```diff
@@ -20,7 +20,10 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-const otelName = "generic"
+const (
+	otelName            = "generic"
+	defaultMaxIdleConns = 2 // default from database/sql
```
bschimke95, line 25: is there a reason you use 5 in the CLI?

bschimke95 (approved):
LGTM, just minor nits/questions

pkg/kine/drivers/sqlite/sqlite_test.go
```diff
@@ -51,7 +53,13 @@ func TestMigration(t *testing.T) {
 	}
 
 	ctx := context.Background()
-	if _, err := sqlite.New(ctx, dbPath); err != nil {
+	connPoolConfig := generic.ConnectionPoolConfig{
+		MaxIdle:     5,
+		MaxOpen:     5,
+		MaxLifetime: 60 * time.Second,
+		MaxIdleTime: 60 * time.Second,
```
marco6, line 60: In the CLI, the default for `MaxIdleTime` is `0*time.Second`. Can we use that here as well?

pkg/server/server.go
```diff
@@ -69,6 +69,10 @@ func New(
 	admissionControlPolicy string,
 	admissionControlPolicyLimitMaxConcurrentTxn int64,
 	admissionControlOnlyWriteQueries bool,
+	datastoreMaxIdleConnections int,
```
marco6, line 72: I think this method was already accepting too many arguments. Can we have an option struct instead? Something like `ServerOptions` or even just `Options`?

marco6 (changes requested):
Nice work on making this more configurable!

I see that this affects performance quite a bit, especially for leases. Can we try to understand why? I just want to make sure that a `MaxLifetime` does not impact with the correctness of the queries (i.e. the connection expires and we miss some updates while reconnecting).

Also, couple of nitpicks I noticed while reading.

pkg/kine/drivers/generic/generic.go
louiseschmidtgen: Five is our current default value.

cmd/root.go
louiseschmidtgen: I thought this could be a little long but I think you're right this is more precise.

louiseschmidtgen: Went with the struct

pkg/kine/drivers/generic/generic.go
louiseschmidtgen: It probably helps to bump this number for large clusters. But that should be possible on bootstrap through this change now.

pkg/kine/drivers/sqlite/sqlite_test.go
louiseschmidtgen: Yes, good catch!

pkg/server/server.go
louiseschmidtgen: Agreed, using the struct from generic now.

louiseschmidtgen:
@marco6 Regarding the benchmark: The lease test calls `leaseGrant` which in return does no backend calls and only returns a response. I am not sure why the `network-bytes-written/op` is increased, however I think the total amount is negligible compared to the other operations calls' total  `network-bytes-written/op`. What do you think?

pkg/kine/drivers/generic/generic.go
```diff
@@ -162,10 +165,32 @@ type Generic struct {
...
+func configureConnectionPooling(connPoolConfig ConnectionPoolConfig, db *sql.DB) {
+	// behavior of database/sql - zero means defaultMaxIdleConns; negative means 0
+	if connPoolConfig.MaxIdle < 0 {
+		connPoolConfig.MaxIdle = 0
+	} else if connPoolConfig.MaxIdle == 0 {
+		connPoolConfig.MaxIdle = defaultMaxIdleConns
+	}
+
+	logrus.Infof(
+		"Configuring database connection pooling: maxIdleConns=%d, maxOpenConns=%d, connMaxLifetime=%s, connMaxIdleTime=%s ",
```
marco6, line 184: Am I wrong or this format string is not quite right? `%s`  should be `%v` for `time.Duration`?

pkg/kine/drivers/generic/generic.go
```diff
@@ -200,7 +225,7 @@ func openAndTest(driverName, dataSourceName string) (*sql.DB, error) {
 	return db, nil
 }
 
-func Open(ctx context.Context, driverName, dataSourceName string, paramCharacter string, numbered bool) (*Generic, error) {
+func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig ConnectionPoolConfig, paramCharacter string, numbered bool) (*Generic, error) {
```
marco6, line 228: While it doesn't make a lot of difference here, in general options and configurations are passed as pointers in Go. 

Can we use a pointer here (and in all other usages of this struct?)

marco6 (approved):
Nice work! :tada:  There are still a couple of nitpicks, but I'm pre-approving this so you can merge as soon as they are fixed!

pkg/kine/drivers/generic/generic.go
louiseschmidtgen: Thanks for catching this!

PR #159, opened 2024-08-14
Timeout long running db queries

louiseschmidtgen:
## Description
This PR catches long running db queries and cancels them. 

- read queries pass through the `query` function
- write queries pass through the `execute` function
- and a batch compaction (1000rows) passes through `tryCompact` (timeout is on each batch)

All db calls made in these functions are context aware and can be cancelled.


louiseschmidtgen:
The timeout can somehow cause an error in the poll loop which is part of a long running go routine that starts with kine and ends with it.

louiseschmidtgen:
Cancelling long running queries has the unwanted side-effect of cancelling the long-running kine go-routine which in effect stops kine.

PR #161, opened 2024-08-27
After query timeout in watcher's poll loop

louiseschmidtgen:
## Description
This PR adds a timeout on long running after queries which are part of the watcher's poll loop. 
The timeout is configurable using the `poll-after-timeout` flag.

## Context
This PR addresses issues raised in microk8s where after the leader node is removed from the cluster the remaining nodes also go into `NotReady` state for ~20 minutes. This timeout helps the responsiveness of the cluster after loosing its leader.

## Note to reviewers
The largest part of this code change is passing the timeout arg to the poll loop.
Look at files in this order: root.go -> server.go -> endpoint.go -> sqlite.go/ dqlite.go -> generic.go -> sql.go

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -37,6 +37,7 @@ var (
 		admissionControlPolicy   string
 		acpLimitMaxConcurrentTxn int64
 		acpOnlyWriteQueries      bool
+		pollAfterTimeout         time.Duration
 	}
 
 	rootCmd = &cobra.Command{
@@ -79,6 +80,7 @@ var (
 				rootCmdOpts.admissionControlPolicy,
 				rootCmdOpts.acpLimitMaxConcurrentTxn,
 				rootCmdOpts.acpOnlyWriteQueries,
+				rootCmdOpts.pollAfterTimeout,
 			)
 			if err != nil {
 				logrus.WithError(err).Fatal("Failed to create server")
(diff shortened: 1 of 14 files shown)


cmd/root.go
```diff
@@ -141,4 +143,5 @@ func init() {
 	// TODO(MK-1408): This value is highly dependend on underlying hardware, thus making the default value a bit useless. The linked card will implement a dynamic way to set this value.
 	rootCmd.Flags().Int64Var(&rootCmdOpts.acpLimitMaxConcurrentTxn, "admission-control-policy-limit-max-concurrent-transactions", 300, "Maximum number of transactions that are allowed to run concurrently. Transactions will not be admitted after the limit is reached.")
 	rootCmd.Flags().BoolVar(&rootCmdOpts.acpOnlyWriteQueries, "admission-control-only-for-write-queries", false, "If set, admission control will only be applied to write queries.")
+	rootCmd.Flags().DurationVar(&rootCmdOpts.pollAfterTimeout, "poll-after-timeout", 20*time.Second, "Timeout on the after query during the poll loop. If timeout is reached, the poll loop will be re-triggered")
```
berkayoz, line 146:
```suggestion
	rootCmd.Flags().DurationVar(&rootCmdOpts.pollAfterTimeout, "poll-events-timeout", 20*time.Second, "Timeout for querying events in the watch poll loop. If timeout is reached, the loop will be re-triggered")
```

Can we rename the parameters to be more descriptive? The `After` keyword does not mean anything to normal users.

pkg/server/server.go
```diff
@@ -273,6 +274,7 @@ func New(
 
 	kineConfig.Listener = listen
 	kineConfig.Endpoint = fmt.Sprintf("dqlite://k8s?%s", params.Encode())
+	kineConfig.PollAfterTimeout = pollAfterTimeout
```
berkayoz, line 277: I think we should update to use our [existing approach](https://github.com/canonical/k8s-dqlite/blob/ca5713a2f267ef2a8b30f165f7675b36cec24ba0/pkg/server/server.go#L264-L278) of passing parameters through the driver string.

berkayoz (changes requested):
Great work, left some comments.

cmd/root.go
louiseschmidtgen: Great point, thanks Berkay!

bschimke95 (approved):
LGTM

PR #162, opened 2024-08-28
Configurable watch query timeout

louiseschmidtgen:
## Description
This PR adds a timeout on long running after queries which are part of the watcher's poll loop. 
The timeout is configurable using the `watch-query-timeout` flag.

## Context
This PR addresses issues raised in microk8s where after the leader node is removed from the cluster the remaining nodes also go into `NotReady` state for ~20 minutes. This timeout helps the responsiveness of the cluster after loosing its leader.

## Note to reviewers
The largest part of this code change is passing the timeout arg to the poll loop.
Look at files in this order: root.go -> server.go (params encoded in driver string) -> sqlite.go (parsed, set on dialect which is generic)-> generic.go (returns timeout) -> sql.go (watcher poll loop)

### DIFF
--- a/pkg/kine/logstructured/sqllog/sql.go
+++ b/pkg/kine/logstructured/sqllog/sql.go
@@ -404,7 +404,6 @@ func (s *SQLLog) poll(result chan interface{}, pollStart int64) {
 
 		rows, err := s.d.After(watchCtx, last, 500)
 		if err != nil {
-			logrus.Errorf("fail to list latest changes: %v", err)
 			if !errors.Is(err, context.DeadlineExceeded) {
 				logrus.Errorf("fail to list latest changes: %v", err)
 			}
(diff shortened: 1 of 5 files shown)


pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -397,10 +399,15 @@ func (s *SQLLog) poll(result chan interface{}, pollStart int64) {
 			}
 		}
 		waitForMore = true
+		watchCtx, cancel := context.WithTimeout(s.ctx, s.d.GetWatchQueryTimeout())
+		defer cancel()
 
-		rows, err := s.d.After(s.ctx, last, 500)
+		rows, err := s.d.After(watchCtx, last, 500)
 		if err != nil {
 			logrus.Errorf("fail to list latest changes: %v", err)
```
bschimke95, line 407: nit: remove this. This should only be printed if error != context.DeadlineExceeded

bschimke95 (approved):
LGTM, minor comment

PR #164, opened 2024-08-29
Fix list benchmark and notify CI on error

marco6:
This PR fixes a problem in the List benchmark. It also makes sure that the CI pipeline for benchmarks fail if an error occurs (it was ignoring the error before).


louiseschmidtgen (approved):
LGTM

louiseschmidtgen (approved):
We are good to merge, running against master fails since master benchmark CI was broken silently before.

PR #165, opened 2024-09-02
Add docs on k8s-dqlite versioning

louiseschmidtgen:
## Description
These docs aim to capture the relationship between k8s version and k8s-dqlite version.


bschimke95 (approved):
Looks good, thanks! I think you could just put that into the root docs folder. This might be interesting info not just for devs.

PR #166, opened 2024-09-02
Remove unused GetRevision query

marco6:
`GetRevision` method (and the associated query) were used in the compaction only. Now no code calls that function.

As such, I removed it.


bschimke95 (approved):
LGTM, good catch!

PR #167, opened 2024-09-03
[WIP] Make benchmark time more accurate

louiseschmidtgen:
## DRAFT
Reduces time counting towards validation or other things and reduces the time to the actual query that we would like to benchmark.
Also removes a piece of dead code.


louiseschmidtgen:
Note that calling lease grant doesn't do anything but return a response to k8s, leases and ttl are handled by a long running go routine in the kine layer: [pkg/kine/logstructured/logstructured.go](https://github.com/canonical/k8s-dqlite/blob/master/pkg/kine/logstructured/logstructured.go#L383).
We can probably remove this test. @marco6 what do you think?

PR #168, opened 2024-09-04
Code cleanup remove lease test and dead code

louiseschmidtgen:
## Description 

Note that calling lease grant doesn't do anything but return a response to k8s, leases and ttl are handled by a long running go routine in the kine layer: [pkg/kine/logstructured/logstructured.go](https://github.com/canonical/k8s-dqlite/blob/master/pkg/kine/logstructured/logstructured.go#L383).
Removing the benchmark test because it is not measuring any lease performance.

The function `updateKey` is not used and can be removed.

### DIFF
--- a/test/lease_test.go
+++ b/test/lease_test.go
@@ -79,30 +79,3 @@ func TestLease(t *testing.T) {
 		})
 	}
 }
-
-// BenchmarkLease is a benchmark for the lease operation.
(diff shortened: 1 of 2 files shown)


bschimke95 (approved):
LGTM

marco6 (approved):

PR #169, opened 2024-09-04
Update versioning docs

louiseschmidtgen:
## Description
Updating versioning docs since MK 1.28 will be pointed at v1.1.11 as well now.


bschimke95 (approved):
LGTM, so no special 1.28 branch anymore? How come?

louiseschmidtgen:
> LGTM, so no special 1.28 branch anymore? How come?

We could not get the snap to pass the core addons upgrade test so we assume there is a bug-fix somewhere between the old version and the newer v1.1.11 version of k8s-dqlite. So we point it to the newer version.

PR #170, opened 2024-09-05
Update Query

louiseschmidtgen:
## Description
This PR creates a separate query for update instead of calling the Append method which does an insert. 
Upon a successful update the response returns the revision of the key.
An unsuccessful update can have the following reasons:
- Key does not exist
- Key is deleted
- We are asked to update a previous revision of the key
In those cases we try to get the latest revision of the key and return it in the response.

Note: We no longer return preRev with a successful update response this was only needed on migration from etcd v2 to etcd v3.

## Performance
This Query can be optimized when we are able to do a schema change and can put an index on deleted.

### DIFF
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -554,8 +554,7 @@ func (d *Generic) Update(ctx context.Context, key string, value []byte, preRev, 
 	if insertCount, err := result.RowsAffected(); err != nil {
 		return 0, err
 	} else if insertCount == 0 {
-		return 0, server.ErrKeyExists
-		// FIXME: Wrong Error
+		return 0, server.ErrRevNotFound
 	}
 	return result.LastInsertId()
 }
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -534,7 +534,7 @@ func (d *Generic) Create(ctx context.Context, key string, value []byte, ttl int6
 	}
 	return result.LastInsertId()
 }
-func (d *Generic) Update(ctx context.Context, key string, value []byte, preRev, ttl int64) (rev int64, err error) {
+func (d *Generic) Update(ctx context.Context, key string, value []byte, preRev, ttl int64) (rev int64, updated bool, err error) {
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Update", otelName))
 	defer func() {
 		if err != nil {
@@ -549,14 +549,15 @@ func (d *Generic) Update(ctx context.Context, key string, value []byte, preRev, 
 	result, err := d.execute(ctx, "update_sql", d.UpdateSQL, key, ttl, value, key, preRev)
 	if err != nil {
 		logrus.WithError(err).Error("failed to update key")
-		return 0, err
+		return 0, false, err
 	}
 	if insertCount, err := result.RowsAffected(); err != nil {
-		return 0, err
+		return 0, false, err
 	} else if insertCount == 0 {
-		return 0, server.ErrRevNotFound
+		return 0, false, nil
 	}
-	return result.LastInsertId()
+	rev, err = result.LastInsertId()
+	return rev, true, err
 }
 
 // Compact compacts the database up to the revision provided in the method's call.
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -149,6 +149,7 @@ type Generic struct {
 	FillSQL               string
 	InsertLastInsertIDSQL string
 	CreateSQL             string
+	UpdateSQL             string
 	GetSizeSQL            string
 	Retry                 ErrRetry
 	TranslateErr          TranslateErr
@@ -311,6 +312,24 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 				WHERE name = ?
 			) maxkv
 			WHERE maxkv.deleted = 1 OR id IS NULL`, paramCharacter, numbered),
+
+		UpdateSQL: q(`
+			INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value)
+			SELECT 
+				? AS name,
+				0 AS created,
+				0 AS deleted,
+				create_revision,
+				prev_revision,
+				? AS lease,
+				? AS value,
+				value AS old_value
+			 FROM(
+				SELECT MAX(id) AS id, deleted, create_revision, prev_revision, value
+				FROM kine
+				WHERE name = ?
+			) maxkv
+			WHERE maxkv.deleted = 0 AND maxkv.id = ?`, paramCharacter, numbered),
 
 		FillSQL: q(`INSERT INTO kine(id, name, created, deleted, create_revision, prev_revision, lease, value, old_value)
 			VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)`, paramCharacter, numbered),
@@ -512,6 +531,31 @@ func (d *Generic) Create(ctx context.Context, key string, value []byte, ttl int6
 		return 0, err
 	} else if insertCount == 0 {
 		return 0, server.ErrKeyExists
+	}
+	return result.LastInsertId()
+}
+func (d *Generic) Update(ctx context.Context, key string, value []byte, preRev, ttl int64) (rev int64, err error) {
+	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Update", otelName))
+	defer func() {
+		if err != nil {
+			if d.TranslateErr != nil {
+				err = d.TranslateErr(err)
+			}
+			span.RecordError(err)
+		}
+		span.End()
+	}()
+
+	result, err := d.execute(ctx, "update_sql", d.UpdateSQL, key, ttl, value, key, preRev)
+	if err != nil {
+		logrus.WithError(err).Error("failed to update key")
+		return 0, err
+	}
+	if insertCount, err := result.RowsAffected(); err != nil {
+		return 0, err
+	} else if insertCount == 0 {
+		return 0, server.ErrKeyExists
+		// FIXME: Wrong Error
 	}
 	return result.LastInsertId()
 }
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -320,13 +320,13 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 				0 AS created,
 				0 AS deleted,
 				create_revision,
-				MAX(id) AS prev_revision,
+				id AS prev_revision,
 				? AS lease,
 				? AS value,
 				value AS old_value
-			FROM kine WHERE name = ?
-			GROUP BY name
-			HAVING deleted = 0 AND id = ?`, paramCharacter, numbered),
+			FROM kine WHERE id = (SELECT MAX(id) FROM kine WHERE name = ?)
+    			AND deleted = 0
+    			AND id = ?`, paramCharacter, numbered),
 
 		FillSQL: q(`INSERT INTO kine(id, name, created, deleted, create_revision, prev_revision, lease, value, old_value)
 			VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)`, paramCharacter, numbered),
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -320,16 +320,12 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 				0 AS created,
 				0 AS deleted,
 				create_revision,
-				id AS prev_revision,
+				MAX(id) AS prev_revision,
 				? AS lease,
 				? AS value,
 				value AS old_value
-			 FROM(
-				SELECT MAX(id) AS id, deleted, create_revision, prev_revision, value
-				FROM kine
-				WHERE name = ?
-			) maxkv
-			WHERE maxkv.deleted = 0 AND maxkv.id = ?`, paramCharacter, numbered),
+			FROM kine WHERE name = ?
+			HAVING deleted = 0 AND id = ?`, paramCharacter, numbered),
 
 		FillSQL: q(`INSERT INTO kine(id, name, created, deleted, create_revision, prev_revision, lease, value, old_value)
 			VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)`, paramCharacter, numbered),
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -320,7 +320,7 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 				0 AS created,
 				0 AS deleted,
 				create_revision,
-				prev_revision,
+				id AS prev_revision,
 				? AS lease,
 				? AS value,
 				value AS old_value
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -325,6 +325,7 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 				? AS value,
 				value AS old_value
 			FROM kine WHERE name = ?
+			GROUP BY name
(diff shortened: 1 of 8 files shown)


marco6 (changes requested):
Thanks for this Louise! Nearly there

pkg/kine/server/update.go
```diff
@@ -46,26 +46,26 @@ func (l *LimitedServer) update(ctx context.Context, rev int64, key string, value
...
 	if rev == 0 {
 		rev, err = l.backend.Create(ctx, key, value, lease)
-		ok = true
-
-		span.SetAttributes(
-			attribute.Int64("revision", rev),
-			attribute.Bool("ok", ok),
-		)
+		if err == ErrKeyExists {
+			updated = false
```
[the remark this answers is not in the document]
louiseschmidtgen: So this would mean there is a case where the update request is asked to update a rev 0, the create returns rev 0 and a nil error? I don't think it is possible or when does this happen?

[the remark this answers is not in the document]
marco6: A create is a transaction like this:

```
if mod_revision(key) == 0
then put(key, value)
else do nothing
```

In the `else` branch we are returning a `ResponseRange` (and performing a `get`) instead of nothing.

In the code the path is:
 - we call `l.backend.Create` and it returns `0, ErrKeyExists` (so `rev = 0` here) as the key is already there;
 - then we do `if err == ErrKeyExists { updated = false; err = nil }` (line 49, 50, 51) which is correct as it is expected by the transaction code that the row could already be there.
 
 However, in this case we reach the `else` branch on line `78` which performs a `get` and returns the prev value. This was not required by this transaction and should really not be done.
 
Reading this again, I don't think adding a simple `if rev != 0` is the solution as that would prevent the `update` path (which does require the prev value) from getting the row. Maybe we should not overwrite `rev`? (it is in general a bad idea to overwrite parameters...)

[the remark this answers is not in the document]
louiseschmidtgen: I think the easiest is to return the response right there that will be easier to follow than adding another case in the later logic. Have a look please and let me know what you think :)

louiseschmidtgen:
Note:
```
sqlite> INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value)
                        SELECT 
                                ? AS name,
                                0 AS created,
                                0 AS deleted,
                                create_revision,
                                MAX(id) AS prev_revision,
                                ? AS lease,
                                ? AS value,
                                value AS old_value
                        FROM kine WHERE name = ?
                        HAVING deleted = 0 AND id = ?;
QUERY PLAN
`--SEARCH kine USING INDEX kine_name_index (name=?)
```

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -565,6 +566,19 @@ func (s *SQLLog) Create(ctx context.Context, key string, value []byte, lease int
 	return rev, nil
 }
 
+func (s *SQLLog) Update(ctx context.Context, key string, value []byte, prevRev, lease int64) (rev int64, updated bool, err error) {
+	rev, updated, err = s.d.Update(ctx, key, value, prevRev, lease)
+	if err != nil {
+		return 0, false, err
+	}
+
+	select {
```
HomayoonAlimohammadi, line 575: maybe we can refactor this select blcok into a `s.maybeNotify(rev)` or `s.tryNotify(rev)`?

pkg/kine/server/update.go
```diff
@@ -46,26 +46,28 @@ func (l *LimitedServer) update(ctx context.Context, rev int64, key string, value
 
 	if rev == 0 {
 		rev, err = l.backend.Create(ctx, key, value, lease)
-		ok = true
-
-		span.SetAttributes(
-			attribute.Int64("revision", rev),
-			attribute.Bool("ok", ok),
-		)
+		if err == ErrKeyExists {
```
HomayoonAlimohammadi, line 49: let's use `errors.Is()`, because: https://go.dev/play/p/3Uo9WyhnQeH

pkg/kine/drivers/generic/generic.go
```diff
@@ -515,6 +531,31 @@ func (d *Generic) Create(ctx context.Context, key string, value []byte, ttl int6
...
+		span.End()
+	}()
+
+	result, err := d.execute(ctx, "update_sql", d.UpdateSQL, key, ttl, value, key, preRev)
+	if err != nil {
+		logrus.WithError(err).Error("failed to update key")
+		return 0, false, err
+	}
+	if insertCount, err := result.RowsAffected(); err != nil {
+		return 0, false, err
```
HomayoonAlimohammadi, line 552: maybe wrap?

pkg/kine/drivers/generic/generic.go
```diff
@@ -515,6 +531,31 @@ func (d *Generic) Create(ctx context.Context, key string, value []byte, ttl int6
...
+	} else if insertCount == 0 {
+		return 0, false, nil
+	}
+	rev, err = result.LastInsertId()
+	return rev, true, err
```
HomayoonAlimohammadi, line 557: I would check for err and wrap if it's not nil

pkg/kine/logstructured/sqllog/sql.go
```diff
@@ -565,6 +566,19 @@ func (s *SQLLog) Create(ctx context.Context, key string, value []byte, lease int
 	return rev, nil
 }
 
+func (s *SQLLog) Update(ctx context.Context, key string, value []byte, prevRev, lease int64) (rev int64, updated bool, err error) {
+	rev, updated, err = s.d.Update(ctx, key, value, prevRev, lease)
+	if err != nil {
+		return 0, false, err
```
HomayoonAlimohammadi, line 572: maybe wrap?

pkg/kine/server/update.go
```diff
@@ -76,6 +78,10 @@ func (l *LimitedServer) update(ctx context.Context, rev int64, key string, value
 			},
 		}
 	} else {
+		rev, kv, err = l.backend.Get(ctx, key, "", 1, rev)
+		if err != nil {
+			return nil, err
```
HomayoonAlimohammadi, line 83: maybe wrap?

test/update_test.go
```diff
@@ -47,6 +33,21 @@ func TestUpdate(t *testing.T) {
...
+
+				createKey(ctx, g, kine.client, "createExistingKey", "testValue1")
+
+				resp, err := kine.client.Txn(ctx).
+					If(clientv3.Compare(clientv3.ModRevision("createExistingKey"), "=", 0)).
+					Then(clientv3.OpPut("createExistingKey", "testValue1")).
+					Else(clientv3.OpGet("createExistingKey", clientv3.WithRange(""))).
+					Commit()
+
+				g.Expect(err).To(BeNil())
```
HomayoonAlimohammadi, line 47: I think this is a bit easier to read
```suggestion
				g.Expect(err).ToNot(HaveOccurred())
```

test/update_test.go
```diff
@@ -65,6 +66,39 @@ func TestUpdate(t *testing.T) {
...
+			t.Run("UpdateNotExistingFails", func(t *testing.T) {
+				g := NewWithT(t)
+
+				resp, err := kine.client.Txn(ctx).
+					If(clientv3.Compare(clientv3.ModRevision("updateNotExistingKey"), "=", 1)).
+					Then(clientv3.OpPut("updateNotExistingKey", "testValue3")).
+					Else(clientv3.OpGet("updateNotExistingKey", clientv3.WithRange(""))).
+					Commit()
+
+				g.Expect(err).To(BeNil())
```
HomayoonAlimohammadi, line 79: also here
```suggestion
				g.Expect(err).ToNot(HaveOccurred())
```

test/update_test.go
```diff
@@ -65,6 +66,39 @@ func TestUpdate(t *testing.T) {
...
+				createKey(ctx, g, kine.client, "updateDeletedKey", "testValue4")
+				lastModRev := deleteKey(ctx, g, kine.client, "updateDeletedKey")
+
+				resp, err := kine.client.Txn(ctx).
+					If(clientv3.Compare(clientv3.ModRevision("updateDeletedKey"), "=", lastModRev)).
+					Then(clientv3.OpPut("updateDeletedKey", "testValue4")).
+					Else(clientv3.OpGet("updateDeletedKey", clientv3.WithRange(""))).
+					Commit()
+
+				g.Expect(err).To(BeNil())
```
HomayoonAlimohammadi, line 97: and here:
```suggestion
				g.Expect(err).ToNot(HaveOccurred())
```

HomayoonAlimohammadi (commented):
Great work @louiseschmidtgen! Thanks a lot! left some minor comments as open discussions. feel free to ignore them if you don't like them.

pkg/kine/server/update.go
marco6: While I in general agree with wrapping an error, I would disagree with this one alone. We would need that only if wrapping and in this case, I think we shouldn't.

This kind of errors are part of the contract between the abstraction and part of the flow: wrapping it in a user-friendly one would only make you allocate to discard it immediately when you are handling it. Then why doing it?

louiseschmidtgen:
Depends on https://github.com/canonical/k8s-dqlite/pull/171

pkg/kine/logstructured/sqllog/sql.go
louiseschmidtgen: We're hoping to remove this implementation when improving the watchers. 
Since we have this call in a few places we could consider making it a separate function.
FYI, currently this wakes up the watcher poll loop to check for the change to the database.

marco6 (approved):
Very nice! A brief summary of the performance-related metrics here.

Some metrics changes were expected like the reduction of the amount of data going through the network:

```
                                        │          Before          │                      After                      │
                                        │ network-bytes-read/op │ network-bytes-read/op  vs base               │
Update/1-workers/dqlite-4                           564.6 ±  0%             262.7 ±  0%  -53.47% (p=0.001 n=7)
Update/2-workers/dqlite-4                           664.6 ±  1%             262.8 ±  0%  -60.46% (p=0.001 n=7)
Update/4-workers/dqlite-4                           709.2 ±  1%             263.2 ±  0%  -62.89% (p=0.001 n=7)
Update/8-workers/dqlite-4                           707.8 ±  1%             262.9 ±  0%  -62.86% (p=0.001 n=7)
Update/16-workers/dqlite-4                          700.3 ±  1%             264.6 ±  1%  -62.22% (p=0.001 n=7)

                                        │ network-bytes-written/op │ network-bytes-written/op  vs base               │
Update/1-workers/dqlite-4                              263.4 ±  0%                136.4 ±  0%  -48.22% (p=0.001 n=7)
Update/2-workers/dqlite-4                              640.9 ±  3%                136.4 ±  0%  -78.72% (p=0.001 n=7)
Update/4-workers/dqlite-4                              797.2 ±  3%                136.5 ±  0%  -82.88% (p=0.001 n=7)
Update/8-workers/dqlite-4                              853.1 ±  3%                136.5 ±  0%  -84.00% (p=0.001 n=7)
Update/16-workers/dqlite-4                             841.7 ±  2%                137.3 ±  1%  -83.69% (p=0.001 n=7)
```

Clearly, we are not moving the row (and the payload of it) unnecessarily from the leader to the client. This is useful not just because we re using a resource less, but it reduces the CPU usage when TLS is on (less data to encrypt).

The reduction of memory usage was expected as well as we are now exploiting what is in sqlite cache locally instead of making multiple copies of it (both in the leader and in the client):

```
                                        │          Before          │                      After                      │
                                        │     B/op      │     B/op       vs base               │
Update/1-workers/dqlite-4                 27.68Ki ±  0%   21.21Ki ±  0%  -23.35% (p=0.001 n=7)
Update/2-workers/dqlite-4                 60.47Ki ±  3%   21.21Ki ±  0%  -64.92% (p=0.001 n=7)
Update/4-workers/dqlite-4                 79.77Ki ±  4%   21.22Ki ±  0%  -73.39% (p=0.001 n=7)
Update/8-workers/dqlite-4                 82.21Ki ±  3%   21.21Ki ±  0%  -74.20% (p=0.001 n=7)
Update/16-workers/dqlite-4                82.00Ki ±  2%   21.39Ki ±  1%  -73.91% (p=0.001 n=7)

                                        │  allocs/op  │  allocs/op   vs base                 │
Update/1-workers/dqlite-4                 492.0 ±  0%   375.0 ±  0%  -23.78% (p=0.001 n=7)
Update/2-workers/dqlite-4                 533.0 ±  0%   375.0 ±  0%  -29.64% (p=0.001 n=7)
Update/4-workers/dqlite-4                 556.0 ±  1%   376.0 ±  0%  -32.37% (p=0.001 n=7)
Update/8-workers/dqlite-4                 547.0 ±  1%   375.0 ±  1%  -31.44% (p=0.001 n=7)
Update/16-workers/dqlite-4                545.0 ±  1%   378.0 ±  1%  -30.64% (p=0.001 n=7)
```

The nice part of this is that now the memory usage is stable, regardless of the number of workers. This is important.

Another thing we were anticipating is the different distribution in time spent in a write transaction vs a read transaction: clearly, we are spending the reading time as a part of an insert, so the "write" part is bigger:

```
                                        │          Before          │                      After                      │
                                        │ sec-reading/op  │ sec-reading/op  vs base                │
Update/1-workers/dqlite-4                  101.90µ ±  12%    13.50µ ±  21%   -86.75% (p=0.001 n=7)
Update/2-workers/dqlite-4                   89.50µ ±  14%    12.60µ ±  29%   -85.92% (p=0.001 n=7)
Update/4-workers/dqlite-4                   65.20µ ±  24%    11.90µ ±  39%   -81.75% (p=0.001 n=7)
Update/8-workers/dqlite-4                   59.40µ ±  20%    10.20µ ±  60%   -82.83% (p=0.001 n=7)
Update/16-workers/dqlite-4                  53.50µ ±  27%    11.40µ ±  32%   -78.69% (p=0.001 n=7)

                                        │ sec-writing/op │ sec-writing/op  vs base                 │
Update/1-workers/dqlite-4                 23.40µ ± 22%       39.50µ ± 23%  +68.80% (p=0.001 n=7)
Update/2-workers/dqlite-4                 29.40µ ± 27%       41.80µ ± 21%  +42.18% (p=0.002 n=7)
Update/4-workers/dqlite-4                 29.10µ ± 28%       43.80µ ± 18%  +50.52% (p=0.001 n=7)
Update/8-workers/dqlite-4                 28.60µ ± 38%       40.30µ ± 12%  +40.91% (p=0.001 n=7)
Update/16-workers/dqlite-4                27.90µ ± 40%       39.60µ ± 25%  +41.94% (p=0.026 n=7)
```
This is not a big concern for me as:
 - the sum of read + write decreased and it is now quite more stable
 - the time can be still optimized by a rather big margin with a schema change
 - batching transaction would have had the same effect on the previous query as well

Also, the main problem of spending more time in a write transaction is not that we are writing more, as shown by the `page-writes` metrics:

```
                                        │          Before          │                      After                      │
                                        │ page-writes/op │ page-writes/op  vs base                │
Update/1-workers/dqlite-4                   4.192 ± 0%         4.208 ± 0%  +0.38% (p=0.001 n=7)
Update/2-workers/dqlite-4                   4.225 ± 1%         4.242 ± 0%  +0.40% (p=0.008 n=7)
Update/4-workers/dqlite-4                   4.189 ± 1%         4.202 ± 0%       ~ (p=0.105 n=7)
Update/8-workers/dqlite-4                   4.210 ± 1%         4.206 ± 1%       ~ (p=0.927 n=7)
Update/16-workers/dqlite-4                  4.230 ± 1%         4.251 ± 1%       ~ (p=0.165 n=7)
```

But that we are keeping the write lock busy for more time and we might reduce concurrency as such. However, it doesn't seem to be the case as even with many concurrent workers we can keep the peace:

```
                                        │    Before     │                After                │
                                        │    sec/op     │    sec/op     vs base               │
Update/1-workers/dqlite-4                  1.364m ±  7%   1.195m ± 14%  -12.35% (p=0.001 n=7)
Update/2-workers/dqlite-4                  1.259m ± 10%   1.106m ± 10%  -12.15% (p=0.001 n=7)
Update/4-workers/dqlite-4                  1.225m ±  8%   1.004m ± 11%  -18.07% (p=0.001 n=7)
Update/8-workers/dqlite-4                 1262.8µ ±  4%   975.5µ ± 13%  -22.75% (p=0.001 n=7)
Update/16-workers/dqlite-4                 1.256m ±  8%   1.044m ±  4%  -16.86% (p=0.001 n=7)
```

This was a very verbose way to say LGTM. Great work Louise!

PR #172, opened 2024-09-20
Rename concurrent tests for create, delete and update

marco6:
This PR renames the subtests that use multiple workers so that they can be easily filtered by backend type.

As of now, doing `go test -run ^$ -branch "./dqlite" does not select those tests. This PR makes sure that they are all part of the same cathegory.


louiseschmidtgen (approved):
LGTM, thanks!

PR #173, opened 2024-09-20
Use constant concatenation for otel span names

marco6:
This is an experimental PR to understand the impact of small concatenation of strings in the hot path. 

I expect this to be more relevant to one-shot queries like GET/CREATE/DELETE/UPDATE rather than in others like LIST or COMPACT.

In general, reducing pressure on the GC is a good thing.


marco6:
We are not yet to the point where this might make a difference, and rebasing this PR on top of other changes is a wasted effort.

We might re-open something similar in the future, but for now let's close this.

PR #175, opened 2024-09-23
Delete Query

louiseschmidtgen:
# Description

This PR moves the logic for deletion into its own query in generic.

## Query logic for deletion

Happy path:

- The specified revision exists in the database and is not already deleted.
- We insert a row with the revision and mark it as deleted via the column deleted=1.
- the prev value is the last revision's value

Edge cases:

1. The specified revision does not exist in the database.
   - We return success=true, empty KV

2. The specified revision is the latest and is already deleted.
    - We want to return the id of this revision.
    - Return success= false and latest rev

3. The specified revision is not the latest revision.
    - We return the latest rev and kvs, success = false

## Query

This is the delete query:
```sql
INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value)
    SELECT 
            ? AS name,
            0 AS created,
            1 AS deleted,
            CASE 
	          WHEN kine.created THEN id
	          ELSE create_revision
	    END AS create_revision,
            id AS prev_revision,
            lease,
            NULL AS value,
            value AS old_value
    FROM kine WHERE id = (SELECT MAX(id) FROM kine WHERE name = ?)
    AND deleted = 0
    AND id = ?;
```

## Optimization for deletion

EXPLAIN QUERY PLAN:

```
QUERY PLAN
|--SEARCH kine USING INTEGER PRIMARY KEY (rowid=?)
`--SCALAR SUBQUERY 1
   `--SEARCH kine USING COVERING INDEX kine_name_index (name=?)
```

Potential optimization with an additional index on deleted.

## Removing unconditional delete

As of Kubernetes v1.15 we do [conditional deletes](https://github.com/kubernetes/kubernetes/blob/v1.15.10/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L195) instead of [unconditional deletes](https://github.com/kubernetes/kubernetes/blob/v1.14.0/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L191). A conditional delete requires the revision of the key to be deleted to match the revision provided in the delete request whereas an unconditional delete does not check for the revision and just deletes what it finds. K8s-dqlite is not used by deployments on v1.14 Kubernetes and can therefore be removed safely.

### DIFF
--- a/pkg/kine/drivers/generic/admission.go
+++ b/pkg/kine/drivers/generic/admission.go
@@ -28,6 +28,8 @@ const (
 var writeQueries = map[string]bool{
 	"update_compact_sql":        true,
 	"delete_sql":                true,
+	"update_sql":                true,
+	"delete_rev_sql":            true,
 	"fill_sql":                  true,
 	"insert_last_insert_id_sql": true,
 	"insert_sql":                true,
--- a/pkg/kine/drivers/generic/generic.go
+++ b/pkg/kine/drivers/generic/generic.go
@@ -309,7 +309,7 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 				value AS old_value
 			FROM kine WHERE id = (SELECT MAX(id) FROM kine WHERE name = ?)
     			AND deleted = 0
-				AND (id = ? OR ? = 0)`, paramCharacter, numbered),
+				AND id = ?`, paramCharacter, numbered),
 
 		CreateSQL: q(`
 			INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value)
@@ -588,7 +588,7 @@ func (d *Generic) Delete(ctx context.Context, key string, revision int64) (rev i
 	}()
 	span.SetAttributes(attribute.String("key", key))
 
-	result, err := d.execute(ctx, "delete_sql", d.DeleteSQL, key, revision, revision)
+	result, err := d.execute(ctx, "delete_sql", d.DeleteSQL, key, revision)
 	if err != nil {
 		logrus.WithError(err).Error("failed to delete key")
 		return 0, false, err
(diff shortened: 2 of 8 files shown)


marco6 (approved):
From the performance POV, the analysis follows the very same pattern as #170 where we see an increase in write time, but we can keep the pace as shown in `sec/op`:

```
Delete/dqlite/1-workers-4                  1.391m ±  5%    1.150m ±   5%  -17.30% (p=0.001 n=7)
Delete/dqlite/2-workers-4                  1.244m ±  2%    1.027m ±   3%  -17.45% (p=0.001 n=7)
Delete/dqlite/4-workers-4                  1.241m ±  4%    1.042m ±   7%  -16.04% (p=0.001 n=7)
Delete/dqlite/8-workers-4                  1.211m ± 16%    1.015m ±  11%  -16.22% (p=0.001 n=7)
Delete/dqlite/16-workers-4                 1.235m ±  5%    1.082m ±   7%  -12.34% (p=0.001 n=7)
```

There is indeed some increase in pages read, but as it can be seen, it is very stable even with more workers, so I'm really not worried:

```
Delete/dqlite/1-workers-4                   24.00 ±  1%     15.31 ±  1%  -36.21% (p=0.001 n=7)
Delete/dqlite/2-workers-4                   12.83 ±  4%     13.25 ±  0%   +3.27% (p=0.001 n=7)
Delete/dqlite/4-workers-4                   10.62 ±  1%     13.22 ±  1%  +24.48% (p=0.001 n=7)
Delete/dqlite/8-workers-4                   10.96 ±  1%     13.21 ±  1%  +20.53% (p=0.001 n=7)
Delete/dqlite/16-workers-4                  11.02 ±  1%     13.23 ±  1%  +20.05% (p=0.001 n=7)
```

Also, this pattern is in common with what we had with the Update query.

Great work Louise!

PR #176, opened 2024-09-25
Fix delete benchmark

marco6:
The delete benchmark right now is using the path for the "[unconditional delete](https://github.com/kubernetes/kubernetes/blob/v1.14.0/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L191)" which is what k8s 1.14 and previous was using. 

Since v1.15 it is now using a [conditional delete](https://github.com/kubernetes/kubernetes/blob/v1.32.0-alpha.0/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L277) instead.

As such, that is what we should test.


louiseschmidtgen (approved):
Thank you Marco!

PR #177, opened 2024-09-25
Fix watch test

marco6:
The current watch test has inter dependencies from subtests and as such subtests cannot be run individually, but only as a whole.

This PR attempts to overcome this issue and make them independent. 

It also fixes some checks:
 - `KV.Value` in delete event should be `nil`
 - `PrevKV.Key` should be correctly populated
 - `[Prev]KV.CreateRevision` should be correctly populated

These new checks highlighted a small bug in the update query which should now be fixed.


marco6:
Depends on #176

marco6:
Ready as #176 was merged.

PR #179, opened 2024-09-26
Add scorescard action and jira sync configuration

petrutlucian94:
Scorecard performs numerous checks to ensure that the project follows security best practices. We're adding a badge to the project readme, which shows the current score and includes various suggestions.

At the same time, we're configuring an automated job that will create Jira issues for each Github issue.

### DIFF
--- a/.github/.jira_sync_config.yaml
+++ b/.github/.jira_sync_config.yaml
@@ -1,0 +1,35 @@
+settings:
+    # Jira project key to create the issue in
+  jira_project_key: "KU"
+
+  # Dictionary mapping GitHub issue status to Jira issue status
+  status_mapping:
+    opened: Untriaged
+    closed: done
+
+  # (Optional) Jira project components that should be attached to the created issue
+  # Component names are case-sensitive
+  components:
+    - K8s snap
+    - Microk8s snap
+
+  # (Optional) GitHub labels. Only issues with one of those labels will be synchronized.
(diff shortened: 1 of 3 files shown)


louiseschmidtgen (approved):
LGTM, much appreciated! @petrutlucian94

.github/.jira_sync_config.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+  # Dictionary mapping GitHub issue status to Jira issue status
+  status_mapping:
+    opened: Untriaged
+    closed: done
+
+  # (Optional) Jira project components that should be attached to the created issue
+  # Component names are case-sensitive
+  components:
+    - K8s snap
```
berkayoz, line 13:
```suggestion
    - K8s dqlite
```

Maybe it's better to differantiate this

berkayoz (approved):
Great work, one small comment. Thank you!

.github/.jira_sync_config.yaml
louiseschmidtgen: Yes there is no such component. In our cards we rarely specify the component field. We can remove it or add microk8s snap there as well since both use k8s-dqlite.

petrutlucian94: Makes senes, I've added "Microk8s snap" as well.

PR #180, opened 2024-09-30
perf: add query batching

marco6:
All update operations (insert/update/delete) force a fsync on the WAL (and sometimes on the disk because of checkpoints). This limits the number of operations to what the underlying disk can do (usually in the order of hundreds).

Grouping these operations in transactions should dramatically increase the throughput while making the top X%ile latency a bit higher. However, this should be balanced by a lower latency in the bottom X%ile.

*We should test this also on a real cluster.*

This PR also removes the write lock as now there is always only one active writer.

### DIFF
--- a/pkg/database/batched.go
+++ b/pkg/database/batched.go
@@ -1,0 +1,215 @@
+package database
+
+import (
+	"context"
+	"database/sql"
+	"sync"
+)
+
+type Execer interface {
+	ExecContext(ctx context.Context, query string, args ...any) (result sql.Result, err error)
+}
+
+type batchStatus int
+
+const (
+	batchNotStarted batchStatus = iota
+	batchStarted
+	batchRunning
+)
+
+type batchedDb[T Transaction] struct {
+	Wrapped[T]
+
+	mu            sync.Mutex
+	workerContext context.Context
+	stopWorker    func()
+	closed        bool
+
+	cv     sync.Cond
+	status batchStatus
+
+	queue []*batchJob
+	runId int64
+}
+
+func NewBatched[T Transaction](db Wrapped[T]) Interface {
+	workerContext, stopWorker := context.WithCancel(context.Background())
+
+	b := &batchedDb[T]{
+		Wrapped:       db,
+		workerContext: workerContext,
+		stopWorker:    stopWorker,
+	}
+	b.cv.L = &b.mu
+	return b
+}
+
+func (db *batchedDb[T]) BeginTx(ctx context.Context, opts *sql.TxOptions) (Transaction, error) {
+	return db.Wrapped.BeginTx(ctx, opts)
+}
+
+func (b *batchedDb[T]) ExecContext(ctx context.Context, query string, args ...any) (sql.Result, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if b.closed {
+		return nil, errDBClosed
+	}
+
+	runId := b.runId
+	if b.status == batchRunning {
+		// The current run is already taking place.
+		runId++
+	}
+
+	job := &batchJob{
+		ctx:   ctx,
+		query: query,
+		args:  args,
+		runId: runId,
+	}
+	b.queue = append(b.queue, job)
+
+	b.run()
+
+	for job.runId >= b.runId {
+		b.cv.Wait()
+	}
+
+	if job.err != nil {
+		return nil, job.err
+	}
+
+	return job, nil
+}
+
+// run starts the batching job if it has not already been started
+// which will run until queue exaustion. run does not block other
+// goroutine from enqueuing new jobs.
+//
+// It must be called while holding the Batch's mu lock.
+func (b *batchedDb[T]) run() {
+	if b.status == batchNotStarted {
+		b.status = batchStarted
+
+		go func() {
+			b.mu.Lock()
+			defer b.mu.Unlock()
+
+			b.status = batchRunning
+			defer func() { b.status = batchNotStarted }()
+
+			for len(b.queue) > 0 {
+				queue := b.queue
+				b.queue = nil
+
+				b.mu.Unlock()
+				b.execQueue(b.workerContext, queue)
+				b.mu.Lock()
+
+				b.runId++
+				b.cv.Broadcast()
+			}
+		}()
+	}
+}
+
+func (b *batchedDb[T]) execQueue(ctx context.Context, queue []*batchJob) {
+	if len(queue) == 0 {
+		return // This should never happen.
+	}
+	if len(queue) == 1 {
+		// We don't need to address the error here as it will be
+		// handled by the goroutine waiting for this result
+		queue[0].exec(queue[0].ctx, b.Wrapped)
+		return
+	}
+
+	transaction := func() error {
+		tx, err := b.BeginTx(ctx, nil)
+		if err != nil {
+			return err
+		}
+		defer tx.Rollback()
+
+		for _, q := range queue {
+			// In the case of SQLITE_FULL SQLITE_IOERR SQLITE_BUSY SQLITE_NOMEM
+			// we should explicitly rollback the whole transaction. In all the other
+			// cases, we could keep going with other queries. However, it is a bit
+			// unclear to me what to do next though as:
+			//  - SQLITE_FULL, SQLITE_IOERR mean that we have problems with the disk
+			//    so, even retrying the batch will not work. We might throttle the
+			//    max batch size, hoping in a checkpoint?
+			// - SQLITE_BUSY should never happen if we manage to get `IMMEDIATE`
+			//   transactions in. Otherwise it only affects the first statement.
+			// - SQLITE_NOMEM, again, we could throttle here? Call a gc collection?
+			// Given the points above, the code below always rolls back the whole
+			// batch. It might seem inefficient, but it should almost never happen.
+			if err := q.exec(ctx, tx); err != nil {
+				return err
+			}
+		}
+
+		return tx.Commit()
+	}
+
+	if err := transaction(); err != nil {
+		for _, q := range queue {
+			q.err = err
+		}
+	}
+}
+
+type batchJob struct {
+	ctx   context.Context
+	query string
+	args  []any
+
+	runId        int64
+	lastInsertId int64
+	rowsAffected int64
+	err          error
+}
+
+var _ sql.Result = &batchJob{}
+
+func (job *batchJob) exec(ctx context.Context, execer Execer) error {
+	select {
+	case <-job.ctx.Done():
+		job.err = job.ctx.Err()
+		return job.err
+	default:
+		// From this point on, the job is not interruptible anymore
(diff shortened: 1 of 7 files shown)


marco6:
Depends on #181

marco6:
Ready as #181 was merged.

pkg/kine/drivers/generic/batch.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return job, nil
+}
+
+// run starts a batching job which will run until queue exaustion.
+// run does not block other goroutine from enqueuing new jobs.
+//
+// It must be called while holding the mu lock.
+func (b *Batch) run() {
```
louiseschmidtgen, line 83: Do we execute b.run() on every call on execute? Would it make sense to wait until we have a couple of txn or a timer runs out to trigger a batch?

pkg/kine/drivers/generic/batch.go
```diff
@@ -0,0 +1,206 @@
...
+		return // This should never happen.
+	}
+	if len(queue) == 1 {
+		// We don't need to address the error here as it will be
+		// handled by the goroutine waiting for this result
+		queue[0].exec(queue[0].ctx, b.db)
+		return
+	}
+
+	transaction := func() error {
```
louiseschmidtgen, line 120: I would like to see a test for batch.go, if possible testing the error cases.

pkg/kine/drivers/generic/generic.go
```diff
@@ -508,7 +500,7 @@ func (d *Generic) Count(ctx context.Context, prefix, startKey string, revision i
 	return rev.Int64, id, err
 }
 
-func (d *Generic) Create(ctx context.Context, key string, value []byte, ttl int64) (rev int64, err error) {
+func (d *Generic) Create(ctx context.Context, key string, value []byte, ttl int64) (rev int64, succeeded bool, err error) {
```
louiseschmidtgen, line 503: Do we need the correction of Create in our v1.2.0? It might be better to have it land in a separate PR.

louiseschmidtgen (commented):
Great work @marco6, took a first look. 
I would like to understand why we have some outliers in our benchmark test, some being as high as +900% through our change. Overall, the benchmark is looking very promising.

pkg/kine/drivers/generic/generic.go
marco6: See #187 .

marco6:
Depends on #186 and #187 .

pkg/kine/drivers/generic/batch.go
marco6: This is executed on every call, but only some will actually spawn a goroutine. The goroutine will stay alive until the queue is empty (which might be for a long time, depending on the status of the queue).

Regarding adding a slight timer to that, it might indeed help. However this is the kind of stuff that I prefer not to do as:
 - we would introduce more latency
 - we don't know how much time to wait (I suspect that the optimal one would be different depending on disk performance)
 
 This one has the first transaction "likely" with a single statement, but then all the other of the burst should be grouped by a time which is the commit time which should decently mimic the timer you are talking about.

pkg/kine/drivers/generic/batch.go
```diff
@@ -0,0 +1,206 @@
...
+		go func() {
+			b.mu.Lock()
+			defer b.mu.Unlock()
+
+			b.status = batchRunning
+			defer func() { b.status = batchNotStarted }()
+
+			for len(b.queue) > 0 {
+				queue := b.queue
+				b.queue = nil
```
TheSignPainter98, line 96: I may be missing something here, but it looks like this line makes the loop only run once? This feels either wrong or fragile to me. Once a batch has started executing, do we want new jobs to be added to its queue, allowing newer items to be merged into the current batch? Although perhaps more performant, this feels like it would make it harder to reason about rollbacks. Should the condition in the `for` be `len(queue) > 0`?

pkg/kine/drivers/generic/batch.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (b *Batch) execQueue(ctx context.Context, queue []*batchJob) {
+	if len(queue) == 0 {
+		return // This should never happen.
+	}
+	if len(queue) == 1 {
+		// We don't need to address the error here as it will be
+		// handled by the goroutine waiting for this result
+		queue[0].exec(queue[0].ctx, b.db)
```
TheSignPainter98, line 116: This feels very fragile! Can we not assign the error into `queue[0].err` as would happen below? I think it could be nicer to keep the responsibility for setting the `err` field here

pkg/kine/drivers/generic/batch.go
```diff
@@ -0,0 +1,206 @@
...
+		// ROLLBACK the whole transaction.
+	}
+
+	var result sql.Result
+	result, bj.err = conn.ExecContext(ctx, bj.query, bj.args...)
+	if bj.err != nil {
+		return bj.err
+	}
+
+	bj.rowsAffected, bj.err = result.RowsAffected()
```
TheSignPainter98, line 185: Assigning into `bj.rowsAffected` even when `bj.err != nil` doesn't feel 'transactional,' so I think it could be cleaner to do this on another line

TheSignPainter98 (commented):
Overall I think this is looking quite nice! Here are a few notes to help polish up the code

marco6:
This PR has been rebased so that the benchmarks results are comparing the current master branch.

pkg/database/prepared.go
```diff
@@ -0,0 +1,168 @@
...
+	return db.underlying.PrepareContext(ctx, query)
+}
+
+func (db *preparedDb) prepare(ctx context.Context, query string) (stmt *sql.Stmt, err error) {
+	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.prepare", otelName))
+	defer func() {
+		span.RecordError(err)
+		span.End()
+	}()
+	span.SetAttributes(attribute.String("query", query))
```
louiseschmidtgen, line 74:
```suggestion
```
Let's remove the query, we don't want to expose any query data to the Otel endpoint.

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
+	"github.com/canonical/k8s-dqlite/pkg/database"
+)
+
+func TestPreparedPrepare(t *testing.T) {
```
louiseschmidtgen, line 11: can we think of a better name?

louiseschmidtgen (commented):
A few small nits and then I'll re-base your branch on the performance test to ensure improvements are made as expected.

marco6:
Given the increased memory requirement from this PR, I think we should wait canonical/dqlite#714.

pkg/database/batched.go
```diff
@@ -0,0 +1,240 @@
...
+	query string
+	args  []any
+
+	runId        int64
+	lastInsertId int64
+	rowsAffected int64
+	err          error
+}
+
+var _ sql.Result = &batchJob{}
```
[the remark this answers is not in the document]
marco6: Given that this is a type assertion, I think it is ok to keep it near the asserted type.

pkg/database/batched.go
```diff
@@ -0,0 +1,240 @@
...
+		defer tx.Rollback()
+
+		for _, q := range queue {
+			// In the case of SQLITE_FULL SQLITE_IOERR SQLITE_BUSY SQLITE_NOMEM
+			// we should explicitly rollback the whole transaction. In all the other
+			// cases, we could keep going with other queries. However, it is a bit
+			// unclear to me what to do next though as:
+			//  - SQLITE_FULL, SQLITE_IOERR mean that we have problems with the disk
+			//    so, even retrying the batch will not work. We might throttle the
+			//    max batch size, hoping in a checkpoint?
```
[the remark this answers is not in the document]
marco6: As discussed some time ago on MM, a compaction would likely use *more* memory (for the WAL pages). It is a bit unclear what to do in these cases, however in general we should never reach this point. As such I think that just reporting the error in the log (and to the user as a failed transaction) is enough.

pkg/database/batched.go
```diff
@@ -0,0 +1,240 @@
...
+			// - SQLITE_BUSY should never happen if we manage to get `IMMEDIATE`
```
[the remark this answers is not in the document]
marco6: There cannot be a snapshot in progress if we managed to start a transaction with `IMMEDIATE`. This code however is not doing that.

I think that we would get a `SQLITE_BUSY`, but only on the first query (as after the first query the read transaction is upgraded to a write one and that would hold the write lock preventing the snapshot).

As such I think it is ok to just rollback as we are not really using a lot work (only the opening of the read transaction).

pkg/database/interface.go
```diff
@@ -0,0 +1,28 @@
+package database
+
+import (
+	"context"
+	"database/sql"
+)
+
+type Interface interface {
```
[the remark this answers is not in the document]
marco6: As discussed on MM I am really out of ideas here. `database.Database` sounds very weird. `database.Interface` is not great, but at least it makes sense. It is also a "pattern" I saw elsewhere, both in the standard library and in k8s itself.

I am open to any suggestion though, anything more expressive is welcome.

marco6:
This PR now depends also on #216 which factors out all the code which is not related to batching from this PR. Some of the comments here are in fact about the refactor rather than batching itself and should be probably discussed there.

pkg/database/batched.go
```diff
@@ -0,0 +1,240 @@
...
+
+	if b.closed {
+		return errDBClosed
+	}
+
+	b.closed = true
+	b.stopWorker()
+
+	for b.status != batchNotStarted {
+		b.cv.Wait()
```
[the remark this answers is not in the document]
marco6: This cannot get stuck as there is a timeout in the execution of the queries (i.e. in the runners). Also releasing the lock here would allow to start again the execution, while the code expects to always have a single batch running.

pkg/database/prepared_test.go
marco6: This was solved in #216

pkg/database/interface.go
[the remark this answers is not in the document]
marco6: Addressed in #216

marco6:
I have rebased this on top of master as #216 was merged. The patch now is way smaller.

marco6:
canonical/dqlite#714 was merged. This PR now depends on the Go side version of it: canonical/go-dqlite#361 .

marco6:
canonical/go-dqlite#361 was merged. Now this PR depends on #248

marco6:
#248 was merged. This PR is now finally ready! :tada:

louiseschmidtgen (approved):
LGTM,
Summary from MM discussion planned for a future iteration:
- Consider improving retry-logic for failed batches (SQLITE_BUSY)
- Consider sharing write lock for write and compact batches

PR #184, opened 2024-10-07
Cancel concurrent workflow k8s-snap integration test

louiseschmidtgen:
# Cancel concurrent workflow k8s-snap integration test.
Through this PR we cancel in flight integration tests that are running when newer changes are pushed.


marco6 (approved):
Thanks Louise! This way we won't run these expensive tests when not needed!

PR #186, opened 2024-10-08
Fix sqlite initialization flags

marco6:
This PR changes the way sqlite is initialized, so that it is similar (in terms of guarantees) to what dqlite is doing.

As such, we want:
 - WAL mode (it was journal mode before);
 - synchronous=FULL (dqlite protects commits from power outages once committed)
 - checked foreign_keys

This might be one of the reasons we see sometimes very different results from dqlite and sqlite.


louiseschmidtgen (approved):
LGTM, we can add a debug log for an unknown option.

PR #187, opened 2024-10-08
Make Create path uniform with Update/Delete

marco6:
This PR makes the Create path look the same as the Update and Delete.

This is mostly a "cosmetic" PR, but I think it would be weird to have such a difference, given how similar those methods are.


louiseschmidtgen (approved):
LGTM, thanks for making everything neat in detail!

PR #190, opened 2024-10-15
Remove old append/insert logic

louiseschmidtgen:
## Remove old append/insert logic
This logic was replaced by create, update and delete logic.


marco6 (approved):
Very nice! Thanks Louise!

PR #192, opened 2024-10-17
Use Otel Collector

louiseschmidtgen:
## Otel Collector
Otel has recently made updates to their [collector](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/examples/otel-collector/README.md), allowing the use of the collector for both metrics and traces and to forward them to jaeger and prometheus.

This code alligns with Otel's example set-up for the collector: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/examples/otel-collector


pkg/instrument/docker-compose.yaml
bschimke95, line 1: The `pkg` folder should only contain go source code. Maybe `hack` fits better?

bschimke95 (commented):
One comment but looks good otherwise.

pkg/instrument/docker-compose.yaml
louiseschmidtgen: True, I'll make a new folder under hack named 'otel' to separate it from all the build scripts.

bschimke95 (approved):
LGTM! Great work @louiseschmidtgen

PR #194, opened 2024-10-29
Merge LogStructured in SQLLog

marco6:
This PR merges the functionality from LogStructured into SQLLog and removes LogStructured.

There are many reasons for this change but the most important one is the lack of a real separation between the two structs, making LogStructured a mere translation layer which adds no useful functionality (and actually acts as a limitation, forcing the implementation to be sub-optimal).

The PR is quite big, but each commit takes care of a single aspect of the transition, making it hopefully possible to perform a review.

Wherever made sense the PR also takes the opportunity to make some of the sub-optimal things better both in term of correctness and in terms of memory footprint/speed (for example avoiding some useless copy-paste or some useless memory allocation).

Some part of this PR (namely, the `List`/`Get` query) could be improved/made simpler, however this PR does not attempt to do that, so that it should be compatible with other outstanding PRs.

As a last note, the write path and the indexes *do not change*, so this PR is compatible with all existing versions.


marco6:
Depends on #193

pkg/kine/sqllog/sqllog.go
```diff
@@ -534,9 +627,67 @@ func (s *SQLLog) notifyWatcherPoll(revision int64) {
...
+	return result, nil
+}
+
+func scanKeyValue(rows *sql.Rows) (*server.KeyValue, error) {
+	kv := &server.KeyValue{}
+	var create, delete bool
+	var prevRevision int64
+	var prevValue []byte
+
+	// FIXME eliminare roba inutile
```
louiseschmidtgen, line 663: Love the comments haha.

pkg/kine/sqllog/sqllog.go
```diff
@@ -96,6 +98,12 @@ func (s *SQLLog) Start(ctx context.Context) error {
 		return err
 	}
 
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		s.ttl(ctx)
+	}()
```
louiseschmidtgen, line 107: Was this not started in the start compaction function? Why move this here?

pkg/kine/broadcaster/broadcaster.go
```diff
@@ -5,21 +5,21 @@ import (
 	"sync"
 )
 
-type ConnectFunc func(ctx context.Context) (chan interface{}, error)
+type ConnectFunc[T any] func(ctx context.Context) (chan T, error)
```
louiseschmidtgen, line 8: Nice!

louiseschmidtgen (commented):
Awesome, thanks for splitting this into separate commits this helped tremendously to follow along the changes one-by-one.
A few questions while I was reading through popped up:

pkg/kine/sqllog/logstructured.go
```diff
@@ -127,11 +124,13 @@ func (l *LogStructured) List(ctx context.Context, prefix, startKey string, limit
 
 func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
 	result := make(chan *server.Event)
-	var shouldClose atomic.Bool
```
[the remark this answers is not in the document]
marco6: Yep.

pkg/kine/sqllog/sqllog.go
marco6: Sorry, I'll remove this.

marco6:
I rebased on top of `master`, this way we see the benchmark diff which does not include other changes.

marco6:
Dropped dependency from #193

pkg/kine/sqllog/sqllog.go
louiseschmidtgen: Fair enough, that makes a lot of sense.

louiseschmidtgen (approved):
LGTM, just removing the comment before merging.

pkg/kine/sqllog/sqllog.go
louiseschmidtgen: removed comment

PR #199, opened 2024-11-11
Performance Test workload (2)

louiseschmidtgen:
## Performance Test Workload

There are two load tests:
- three control plane nodes
- single node cluster

On the cluster an api-intensive job is launched using kube-burner.
This test creates configmaps and secrets and deletes them.

While the load test runs, pidstat is used to get metrics on CPU, MEM, IO reads and writes and output those files for each node.
This is the start of such a metrics file:
```
Linux 6.8.0-48-generic (k8s-performance-b07a0d-4) 	11/12/24 	_x86_64_	(8 CPU)

# Time        UID       PID    %usr %system  %guest   %wait    %CPU   CPU  minflt/s  majflt/s     VSZ     RSS   %MEM   kB_rd/s   kB_wr/s kB_ccwr/s iodelay  Command
15:03:17        0      1785    1.98    3.96    0.00    0.00    5.94     4    261.39      0.00 1391316  165368   0.68      0.00    106.93      0.00       0  k8s-dqlite

# Time        UID       PID    %usr %system  %guest   %wait    %CPU   CPU  minflt/s  majflt/s     VSZ     RSS   %MEM   kB_rd/s   kB_wr/s kB_ccwr/s iodelay  Command
15:03:18        0      1785    2.00    1.00    0.00    0.00    3.00     4    143.00      0.00 1391316  165368   0.68      0.00     68.00      0.00       0  k8s-dqlite

# Time        UID       PID    %usr %system  %guest   %wait    %CPU   CPU  minflt/s  majflt/s     VSZ     RSS   %MEM   kB_rd/s   kB_wr/s kB_ccwr/s iodelay  Command
15:03:19        0      1785    1.00    1.00    0.00    0.00    2.00     4     46.00      0.00 1391316  165368   0.68      0.00     20.00      0.00       0  k8s-dqlite

```

This PR should be merged after https://github.com/canonical/k8s-dqlite/pull/197

### Note for reviewers
Popen is necessary to run the metrics collection command in the background.


test/performance/tests/conftest.py
```diff
@@ -187,3 +188,44 @@ def instances(
...
+    Bootstraps the instance with all k8sd features enabled to reduce testing time.
+    """
+    LOG.info("Setup node and enable all features")
+
+    tmp_path = tmp_path_factory.mktemp("data")
+    instance = h.new_instance()
+    snap = next(snap_versions(request))
+    util.setup_k8s_snap(instance, tmp_path, snap)
+
+    bootstrap_config_path = "/root/bootstrap-session.yaml"
```
HomayoonAlimohammadi, line 208: Maybe we can make this configurable for other tests?

test/performance/tests/test_performance.py
```diff
@@ -30,3 +31,91 @@ def test_load_test(instances: List[harness.Instance]):
...
+        )
+
+
+def configure_kube_burner(instance: harness.Instance):
+    """Downloads and sets up `kube-burner` on each instance if it's not already present."""
+    if (
+        not instance.exec(["test", "-f", "/root/kube-burner"], check=False).returncode
+        == 0
+    ):
+        url = "https://github.com/kube-burner/kube-burner/releases/download/v1.2/kube-burner-1.2-Linux-x86_64.tar.gz"
```
HomayoonAlimohammadi, line 97: Maybe we can make the `kube-burner` version configurable?

test/performance/tests/test_performance.py
```diff
@@ -30,3 +31,91 @@ def test_load_test(instances: List[harness.Instance]):
...
+        instance.exec(["wget", url])
+        instance.exec(
+            ["tar", "-zxvf", "kube-burner-1.2-Linux-x86_64.tar.gz", "kube-burner"]
+        )
+        instance.exec(["rm", "kube-burner-1.2-Linux-x86_64.tar.gz"])
+        instance.exec(["chmod", "+x", "/root/kube-burner"])
```
HomayoonAlimohammadi, line 103: I think a new line here can make it read better

test/performance/tests/test_performance.py
```diff
@@ -30,3 +31,91 @@ def test_load_test(instances: List[harness.Instance]):
...
+        instance.exec(["apt-get", "install", "-y", "sysstat"])
+        subprocess = instance.exec_with_popen(
+            [
+                "pidstat",
+                "-druh",
+                "-p",
+                pid,
+                "1",
+                ">",
+                f"/root/{instance.id}_metrics.log",
```
HomayoonAlimohammadi, line 76: nit: Maybe we can extract this naming convention since they're being used in multiple functions.

test/performance/tests/test_performance.py
```diff
@@ -30,3 +31,91 @@ def test_load_test(instances: List[harness.Instance]):
...
+    run_kube_burner(cluster_node)
+    stop_metrics(instances, process_dict)
+    pull_metrics(instances)
+
+
+def test_single_node_load(session_instance: harness.Instance):
+    """Test the performance of a single node cluster with all features enabled."""
+    configure_kube_burner(session_instance)
+    process_dict = collect_metrics([session_instance])
+    run_kube_burner(session_instance)
```
HomayoonAlimohammadi, line 46: Will the kube-burner exits only when it's done pressure testing? Or will it start and continue in the background?
I'm asking this because I want to make sure we're not killing the metrics processes when the bench is still ongoing.

test/performance/tests/test_performance.py
```diff
@@ -30,3 +31,91 @@ def test_load_test(instances: List[harness.Instance]):
...
+    stop_metrics([session_instance], process_dict)
+    pull_metrics([session_instance])
+
+
+def stop_metrics(instances: List[harness.Instance], process_dict: dict):
```
HomayoonAlimohammadi, line 51: Maybe we can move the bench functionality in a dedicated package.

test/performance/tests/test_util/harness/lxd.py
```diff
@@ -160,8 +161,35 @@ def exec(self, instance_id: str, command: list, **kwargs):
             raise HarnessError(f"unknown instance {instance_id}")
 
         LOG.debug("Execute command %s in instance %s", command, instance_id)
+        # Check if the command contains any special shell characters that need special handling
+        special_characters = ["$", ">", "&", "!", "~", "||", "&&", "|"]
```
HomayoonAlimohammadi, line 165: Do we care about other characters as well? e.g. `<`
Also I think `&&` and `||` can be removed, we have single versions of them (AFAIK `char in " ".join(command)` will catch those).

test/performance/tests/test_util/util.py
```diff
@@ -322,3 +331,48 @@ def tracks_least_risk(track: str, arch: str) -> str:
...
+
+def get_default_ip(instance: harness.Instance):
+    # ---
+    # default via 10.42.254.1 dev eth0 proto dhcp src 10.42.254.197 metric 100
+    # ---
+    # Fetching the default IP address from the output, e.g. 10.42.254.197
+    p = instance.exec(
+        ["ip", "-o", "-4", "route", "show", "to", "default"], capture_output=True
+    )
+    return p.stdout.decode().split(" ")[8]
```
HomayoonAlimohammadi, line 355: What if `p` has multiple lines?

test/performance/tests/test_util/util.py
```diff
@@ -322,3 +331,48 @@ def tracks_least_risk(track: str, arch: str) -> str:
...
+
+
+def find_suitable_cidr(parent_cidr: str, excluded_ips: List[str]):
+    net = ipaddress.IPv4Network(parent_cidr, False)
+
+    # Starting from the first IP address from the parent cidr,
+    # we search for a /30 cidr block(4 total ips, 2 available)
+    # that doesn't contain the excluded ips to avoid collisions
+    # /30 because this is the smallest CIDR cilium hands out IPs from
```
HomayoonAlimohammadi, line 364: Maybe we can also put a link to this claim for future travelers.

test/performance/tests/test_util/util.py
```diff
@@ -322,3 +331,48 @@ def tracks_least_risk(track: str, arch: str) -> str:
...
+        contains_excluded = False
+        for excluded in excluded_ips:
+            if ipaddress.ip_address(excluded) in lb_net:
+                contains_excluded = True
+                break
+
+        if contains_excluded:
+            continue
+
+        return str(lb_net)
```
HomayoonAlimohammadi, line 377:
```suggestion
        if not contains_excluded:
            return str(lb_net)
```

test/performance/tests/conftest.py
louiseschmidtgen: This is a good idea for the integration test in the k8s-snap repo. For k8s-dqlite we do a load-test with a snap that has all the features enabled. I highly doubt we will need a different bootstrap config in the future- I would rather keep this simple for now.

test/performance/tests/test_performance.py
louiseschmidtgen: Yes, I would like this to be arch configurable too. Unfortunately, our friends at kube-burner have been inconsistent with their naming across releases, meaning we can only make the entire url configurable instead.

test/performance/tests/test_performance.py
louiseschmidtgen: Since the filename is dynamic, this is likely going to add unnecessary complexity to the code.
In part 3, this filename only shows up twice- we should be fine.

test/performance/tests/test_util/harness/lxd.py
louiseschmidtgen: A little more information here, I've added this if clause as `shlex.join` did not work with any of my commands with special characters.

So now this line checks for any of those special characters and does a regular join on the command bits:
```
any(char in " ".join(command) for char in special_characters)
```

HomayoonAlimohammadi: Gotcha. Thanks!

test/performance/tests/test_util/util.py
louiseschmidtgen: I think the comment should be sufficient here.

test/performance/tests/test_util/util.py
louiseschmidtgen: For now there only exists one route to default in lxd, when this changes we can adjust the code.

test/performance/tests/test_performance.py
louiseschmidtgen: Great question! The kube-burner command runs and blocks whereas we start the metrics subprocess in the background and kill it after the load test is over.

test/performance/tests/test_util/harness/lxd.py
louiseschmidtgen: I removed some duplication and reduced special char to the one I need :)

test/performance/tests/test_util/config.py
```diff
@@ -25,6 +26,12 @@
 # SNAP_NAME is the name of the snap under test.
 SNAP_NAME = os.getenv("TEST_SNAP_NAME") or "k8s"
 
+# KUBE_BURNER_URL is the version of kube-burner to use.
+KUBE_BURNER_URL = (
+    os.getenv("TEST_KUBE_BURNER_VERSION")
+    or "https://github.com/kube-burner/kube-burner/releases/download/v1.2/kube-burner-1.2-Linux-x86_64.tar.gz"
```
addyess, line 32: when does this ever change?  Should we keep it up-to-date?

test/performance/tests/test_util/harness/lxd.py
```diff
@@ -155,13 +156,23 @@ def pull_file(self, instance_id: str, source: str, destination: str):
...
 
-    def exec(self, instance_id: str, command: list, **kwargs):
+    def exec(self, instance_id: str, command: list, background: bool = False, **kwargs):
         if instance_id not in self.instances:
             raise HarnessError(f"unknown instance {instance_id}")
 
         LOG.debug("Execute command %s in instance %s", command, instance_id)
+
+        if ">" in " ".join(command):
+            command_str = " ".join(command)
```
addyess, line 166: hang on... are you looking for a shell redirect here by searching for `>`.  This feels fragile.  I wonder if i can help think of a better way

louiseschmidtgen: The shlex.join doesn't work with special characters so if my command has a ">" in it I am doing a regular join.

addyess (approved):
Tests look like they work with the interesting redirects + background exec calls.  Let's ship it.  We can pull it back together with a library at some point perhaps

test/performance/tests/test_util/config.py
louiseschmidtgen: We could bump the version in the future or overwrite it with a url for a different architecture. 
As discussed with Hue in an earlier comment our friends at kube burner are not consistently naming their release urls so I decided to make the entire url configurable and not just `arch` and `version`.

louiseschmidtgen: Dealing with this in part 3.

PR #200, opened 2024-11-12
Performance Test Parse Metrics (3)

louiseschmidtgen:
## Performance Test Parse Metrics

This PR takes care of parsing the metrics from the load test (single & three-node) and uploading the results to an artifact.
(GH does not allow uploading images to GH comments- this was the original plan).

We generate the following files & directories:

| Folder | Files |
|------------| --------|
| head | head-1-three-node.log, head-2-three-node.log, head-3-three-node.log , head-1-single-node.log |
| base-code | base-code-1-three-node.log, base-code-2-three-node.log, base-code-3-three-node.log , base-code-1-single-node.log 
| v1-1-11 | ... |
| v1-2-0  | ... |

The test creates another directory for the single-node comparison:
| Folder | Files |
|------------| --------|
| single-node| head-1-single-node.log, base-code-1-single-node.log, v1-2-0 -1-single-node.log, v1-1-11-1-single-node.log|

The script creates graphs for cpu, mem, io reads and writes for all files matching a pattern in a directory. 

Here is an example for the head's 3 node cpu performance:
(Kube-burner sleeps for 60 seconds before starting the test)

image

Here is an example for the single-node io writes:
image

These can be retrieved from the CI run's [artifact](https://github.com/canonical/k8s-dqlite/actions/runs/11836102993/artifacts/2187531162).

Depends on: https://github.com/canonical/k8s-dqlite/pull/199

### DIFF
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -6,6 +6,7 @@ on:
   pull_request:
 
 permissions:
+  pull-requests: write
   contents: read
 
 concurrency:
@@ -39,6 +40,10 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "1.22"
+      - name: Set-up R
+        run: |
+          sudo apt install r-base
+          sudo Rscript -e 'install.packages(c("ggplot2", "dplyr", "optparse"), repos="https://cloud.r-project.org")'
       - name: Install lxd
         run: |
           sudo snap refresh lxd --channel 5.21/stable
@@ -57,21 +62,13 @@ jobs:
       - name: Unpack Snap
         run: |
           sudo unsquashfs -d snap-unpack-dir k8s.snap
-      - name: Create snap with k8s-dqlite ${{ github.head_ref }}
+      - name: Create snap with k8s-dqlite ${{ github.head_ref }} code
         run: |
           make static
           sudo cp ./bin/static/k8s-dqlite snap-unpack-dir/bin/k8s-dqlite
           sudo chmod o+r snap-unpack-dir/bin/k8s-dqlite
           sudo mksquashfs snap-unpack-dir head.snap -noappend -comp lzo -no-fragments
-      - name: Run Performance test ${{ github.head_ref }} snap
-        env:
-          TEST_SNAP: ${{ github.workspace }}/head.snap
-          TEST_SUBSTRATE: lxd
-          TEST_LXD_IMAGE: ubuntu:22.04
-          TEST_INSPECTION_REPORTS_DIR: ${{ github.workspace }}/inspection-reports
-        run: |
-          cd test/performance && sg lxd -c 'tox -e performance'
-      - name: Create snap with k8s-dqlite base code
+      - name: Create snap with k8s-dqlite base-code
         run: |
           set -o pipefail
           git fetch origin $BASE_BRANCH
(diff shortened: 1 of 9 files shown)


.github/workflows/performance.yaml
```diff
@@ -0,0 +1,179 @@
...
+          sudo mksquashfs snap-unpack-dir head.snap -noappend -comp lzo -no-fragments
+      - name: Create snap with k8s-dqlite base-code
+        run: |
+          set -o pipefail
+          git fetch origin $BASE_BRANCH
+          git reset --hard $BASE_SHA
+          make static
+          sudo cp ./bin/static/k8s-dqlite snap-unpack-dir/bin/k8s-dqlite
+          sudo chmod o+r snap-unpack-dir/bin/k8s-dqlite
+          sudo mksquashfs snap-unpack-dir base-code.snap -noappend -comp lzo -no-fragments
```
addyess, line 79: 🪄

test/performance/lxd-profile.yaml
```diff
@@ -0,0 +1,105 @@
+description: "LXD profile for Canonical Kubernetes"
```
addyess, line 1: if you need this profile.  can you pull it from https://github.com/canonical/k8s-snap/blob/main/tests/integration/lxd-profile.yaml

test/performance/parse-performance-metrics.R
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env Rscript
```
addyess, line 1: I am not familiar with r-script.  Did you pull this from somewhere else?  Should we be vendoring this into the project or can it be sourced from elsewhere?

test/performance/tests/test_performance.py
```diff
@@ -0,0 +1,122 @@
...
+        not instance.exec(["test", "-f", "/root/kube-burner"], check=False).returncode
+        == 0
+    ):
+        url = "https://github.com/kube-burner/kube-burner/releases/download/v1.2/kube-burner-1.2-Linux-x86_64.tar.gz"
+        instance.exec(["wget", url])
+        instance.exec(
+            ["tar", "-zxvf", "kube-burner-1.2-Linux-x86_64.tar.gz", "kube-burner"]
+        )
+        instance.exec(["rm", "kube-burner-1.2-Linux-x86_64.tar.gz"])
+        instance.exec(["chmod", "+x", "/root/kube-burner"])
```
addyess, line 104: you should prepare for the eventuality this version 1.2 will change or be gone.  Can you make this less static?

addyess (commented):
Most of my comments are around sourcing of various files.  I don't quite know how to review more than the logic bits of the rscript which looks valid enough

test/performance/parse-performance-metrics.R
louiseschmidtgen: Oh, this is my magic, I used to work for R :)

test/performance/tests/test_performance.py
louiseschmidtgen: Good point let me see what I can do, they did change the naming convention of their release tar for later versions...

test/performance/lxd-profile.yaml
louiseschmidtgen: Okay, since I don't expect to configure it in the future, I'm removing the configuration option- we can always add it back later if we need it.

.github/workflows/performance.yaml
```diff
@@ -120,14 +97,75 @@ jobs:
...
+          TEST_SUBSTRATE: lxd
+          TEST_LXD_IMAGE: ubuntu:22.04
+          TEST_INSPECTION_REPORTS_DIR: ${{ github.workspace }}/inspection-reports
+          TEST_METRICS_DIR: ${{ github.workspace }}/test/performance/results/base-code
+          TEST_RUN_NAME: base-code
+        run: |
+          cd test/performance 
+          mkdir -p ./results/base-code
+          sg lxd -c 'tox -e performance'
+      - name: Run Performance test for v1.1.11 snap
```
eaudetcobello, line 124: Out of curiosity, why are you specifically testing v1.1.11? What's special about this version?

louiseschmidtgen: The v1.1.11 is the tag before we introduced the compaction changes (v1.2.0).
So we can compare where we are now with v1.11.1:

image

addyess (approved):
LGTM

PR #206, opened 2024-11-15
Performance test remove tag comparison

louiseschmidtgen:
## Description

Removing tag comparison on the tags, this is unnecessary because we can get the comparison by changing the base-code of a PR.


bschimke95 (approved):
LGTM

PR #212, opened 2024-12-10
Watch Progress Request

louiseschmidtgen:
# Watch Progress Request

This PR adds support for watch progress notifications. Progress notifications ensure api-server's watch cache is consistent with Dqlite while reducing the need for resource-intensive quorum reads from Dqlite.

The api-server builds a watch history on its side based on an initial list and then a watch stream of events (update/create/delete). Reading from this cache can lead to issues from stale reads. This is where the progress notifications come into play: Upon a progress request to k8s-dqlite's watch-server we return the current revision (in other words the revision of the last event received). If the api-server's cache is not out of date then it can serve reads from the cache instead of from Dqlite.

## Support for Upstream Features:

Through progress notifications support, upstream k8s WatchList and ConsistentListFromCache features are enabled.
- ConsistentListFromCache: Enhance Kubernetes API server performance by serving consistent list requests directly from its watch cache, improving scalability and response times. 
- WatchList: Enable support for [streaming initial state of objects in watch requests](https://kubernetes.io/docs/reference/using-api/api-concepts/#streaming-lists).

For more details read [feature-gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/)

## Kine reference

The PR takes inspiration from kine’s watcher PRs:

- https://github.com/k3s-io/kine/pull/238 
- https://github.com/k3s-io/kine/pull/251
- https://github.com/k3s-io/kine/pull/268
- https://github.com/k3s-io/kine/pull/316
- https://github.com/k3s-io/kine/pull/373

Note to reviewer: I would recommend taking a look at each one of these PRs before diving into this PR.

### Emulated Etcd version 

Support for etcd version check that evaluates whether RequestWatchProgress is supported by the current version of etcd endpoint from status request: https://github.com/kubernetes/kubernetes/blob/beb696c2c9467dbc44cbaf35c5a4a3daf0321db3/staging/src/k8s.io/apiserver/pkg/storage/feature/feature_support_checker.go#L157

Backport of https://github.com/k3s-io/kine/pull/316.

The PR bumps the emulated etcd version to signal support of the api-server's `RequestWatchProgress` (v3.4.31+ or v3.5.13+).

### Watch Channels

For this implementation we need two channels for the watchers:
- watchEventsChannel: receives notifications on revisions created, updated and deleted
- progressChannel: progress notifications are sent every 5s unless configured otherwise in the `watch-progress-notify-interval` flag

When a watch is requested on a revision that has been compacted we return a cancel response, along with the message that the key has been compacted and provide the current revision as well as the compact revision.

## Testing
1.32 rc & test_config_propagation passed https://github.com/canonical/k8s-dqlite/actions/runs/12257281237/workflow
1.32 rc & full integration suite https://github.com/canonical/k8s-dqlite/actions/runs/12293052446/job/34304958695
Unit test: requesting and receiving ProgressNotify message


louiseschmidtgen:
The preliminary performance test with 1.31 suggests that the `ConsistentListOnCache` upstream feature is helping to reduce the resource pressure on k8s-dqlite:

cpu_usage_plot

berkayoz (commented):
LGTM overall, I'll do another pass once finalized. We should split these into multiple PRs(one per upstream PR) as it is really hard to read and follow which change correlates to which PR. If we have to do this in a single PR then commenting through the PR on the changed lines to provide context would be appreciated.

cmd/root.go
```diff
@@ -173,6 +177,7 @@ func init() {
 	rootCmd.Flags().BoolVar(&rootCmdOpts.otel, "otel", false, "enable traces endpoint")
 	rootCmd.Flags().StringVar(&rootCmdOpts.otelAddress, "otel-listen", "127.0.0.1:4317", "listen address for OpenTelemetry endpoint")
 	rootCmd.Flags().StringVar(&rootCmdOpts.metricsAddress, "metrics-listen", "127.0.0.1:9042", "listen address for metrics endpoint")
+	rootCmd.Flags().StringVar(&rootCmdOpts.emulatedEtcdVersion, "emulated-etcd-version", "3.5.13", "The emulated etcd version to return on a call to the status endpoint. Defaults to 3.5.13, in order to indicate no support for watch progress notifications yet.")
```
bschimke95, line 180: IIRC then 3.5.12 was the last non supported release. 3.5.13 is enabling this feature.

docs/configuration.md
```diff
@@ -21,6 +21,7 @@ The following configuration options are available listed in a table format:
 | `--otel` | Enable traces endpoint | `false` |
 | `--otel-listen` | The address to listen for OpenTelemetry endpoint | `127.0.0.1:4317` |
 | `--metrics-listen` | The address to listen for metrics endpoint | `127.0.0.1:9042` |
+| `--emulated-etcd-version` | The emulated etcd version to return on a call to the status endpoint. Defaults to 3.5.13, in order to indicate no support for watch progress notifications yet. | `3.5.13` |
```
bschimke95, line 24: same here

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/canonical/k8s-dqlite
 
-go 1.22
+go 1.22.0
+
+toolchain go1.22.10
```
bschimke95, line 5: why is this needed? Are we really relying on this specific patch version to build this?

pkg/kine/server/get.go
```diff
@@ -23,11 +23,9 @@ func (l *LimitedServer) get(ctx context.Context, r *etcdserverpb.RangeRequest) (
 		attribute.Int64("revision", r.Revision),
 	)
 
-	if len(r.RangeEnd) != 0 {
-		return nil, fmt.Errorf("unexpected rangeEnd: want empty, got %s", r.RangeEnd)
-	}
-	if r.Limit != 0 {
-		return nil, fmt.Errorf("unexpected limit: want 0, got %d", r.Limit)
+	if r.Limit != 0 && len(r.RangeEnd) != 0 {
```
louiseschmidtgen, line 26: This addresses the issue `unexpected limit: want 0, got 1` we saw on 1.32 when enabling the WatchList feature.

pkg/kine/server/get.go
```diff
@@ -23,11 +23,9 @@ func (l *LimitedServer) get(ctx context.Context, r *etcdserverpb.RangeRequest) (
...
+		err := fmt.Errorf("invalid combination of rangeEnd and limit, limit should be 0 got %d", r.Limit)
```
bschimke95, line 27: just curious here: This basically enforces no limit on a range request, right?
Why can't we have a range request with a limit which would just return the first X entries?

bschimke95 (commented):
did a first pass - great work @louiseschmidtgen

go.mod
berkayoz: I believe `go mod tidy` started adding this, see https://github.com/golang/go/issues/65847

cmd/root.go
```diff
@@ -32,15 +32,17 @@ var (
 		metricsAddress         string
 		otel                   bool
 		otelAddress            string
+		emulatedEtcdVersion    string
```
louiseschmidtgen, line 35: Re-opening the discussion on the emulatedEtcdVersion flag here @berkayoz, @bschimke95 and @marco6.

It would not make sense for users to drop the version, we may only bump it in the future if some new fancy feature requires it. 

If a user wishes to disable features they should do so in the api-server.
As a consequence, we do not have to make this `emulatedEtcdVersion` configurable.

pkg/kine/server/maintenance.go
```diff
@@ -19,8 +19,9 @@ func (s *KVServerBridge) Status(ctx context.Context, r *etcdserverpb.StatusReque
 		return nil, err
 	}
 	return &etcdserverpb.StatusResponse{
-		Header: &etcdserverpb.ResponseHeader{},
-		DbSize: size,
+		Header:  &etcdserverpb.ResponseHeader{},
+		DbSize:  size,
+		Version: s.emulatedEtcdVersion,
```
louiseschmidtgen, line 24: The emulatedEtcdVersion  is returned on status calls, giving an indication to the api-server whether the watch progress requests are supported or not.

pkg/kine/server/server.go
```diff
@@ -12,19 +13,26 @@ import (
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 )
 
+// The emulated etcd version is returned on a call to the status endpoint. The version 3.5.13, indicates support for the watch progress notifications.
+// See: https://github.com/kubernetes/kubernetes/blob/beb696c2c9467dbc44cbaf35c5a4a3daf0321db3/staging/src/k8s.io/apiserver/pkg/storage/feature/feature_support_checker.go#L157
+const emulatedEtcdVersion = "3.5.13"
```
louiseschmidtgen, line 18: Re-opening the discussion on the emulatedEtcdVersion flag here @berkayoz, @bschimke95 and @marco6.

It would not make sense for users to drop the version, we may only bump it in the future if some new fancy feature requires it.

If a user wishes to disable features they should do so in the api-server.
As a consequence, we do not have to make this emulatedEtcdVersion configurable.

cmd/root.go
louiseschmidtgen: removed: https://github.com/canonical/k8s-dqlite/pull/212#discussion_r1879804010

docs/configuration.md
louiseschmidtgen: removed: https://github.com/canonical/k8s-dqlite/pull/212#discussion_r1879804010

pkg/kine/server/watch.go
```diff
@@ -67,36 +103,79 @@ func (w *watcher) Start(ctx context.Context, r *etcdserverpb.WatchCreateRequest)
...
+			w.Cancel(id, wr.CurrentRevision, wr.CompactRevision, ErrCompacted)
 			return
 		}
-		for events := range watchCh {
-			if len(events) == 0 {
-				continue
+
+		trace := logrus.IsLevelEnabled(logrus.TraceLevel)
+		outer := true
+		for outer {
```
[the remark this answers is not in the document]
louiseschmidtgen: I've added some comments to this block- does that help clarify?
Basically we are waiting on events but if a progress notify request comes we send a progress report whether there have been events or not.

pkg/kine/server/get.go
```diff
@@ -23,13 +23,6 @@ func (l *LimitedServer) get(ctx context.Context, r *etcdserverpb.RangeRequest) (
 		attribute.Int64("revision", r.Revision),
 	)
 
-	if len(r.RangeEnd) != 0 {
-		return nil, fmt.Errorf("unexpected rangeEnd: want empty, got %s", r.RangeEnd)
-	}
-	if r.Limit != 0 {
-		return nil, fmt.Errorf("unexpected limit: want 0, got %d", r.Limit)
-	}
```
louiseschmidtgen, line 31: Previous check (currently also in kine), essentially never executed:
```
	if r.Limit != 0 && len(r.RangeEnd) != 0 {
		return nil, fmt.Errorf("invalid combination of rangeEnd and limit, limit should be 0 got %d", r.Limit)
	}
```
As there is a restriction already in limited.go for this get request:
```
	if len(r.RangeEnd) == 0 {
		return l.get(ctx, r)
	}
```
Zero limit means -> no limit for the request
In all cases we want 1 revision return by this request.

pkg/kine/server/watch.go
```diff
@@ -4,49 +4,78 @@ import (
...
+		}
+		if pr := msg.GetProgressRequest(); pr != nil {
+			w.Progress(ws.Context())
 		}
 	}
 }
 
 type watcher struct {
-	sync.Mutex
+	sync.RWMutex
```
marco6, line 64: Read-write mutexes have quite some overhead and are usually used only when there is a big number of readers and a very small number of writers. Is this there's only 2 cases where the lock is used:
 - to edit the watchers list (add/remove watcher)
 - to send a report.
 
 These two operations are mutual, so I think that this behaves just like a normal mutex, but with overhead.

pkg/kine/server/watch.go
```diff
@@ -136,23 +220,85 @@ func (w *watcher) Cancel(watchID int64, err error) {
...
+		delete(w.watches, id)
 	}
 	w.Unlock()
 	w.wg.Wait()
 }
+
+// Progress sends a progress report if all watchers are synced.
+// Ref: https://github.com/etcd-io/etcd/blob/v3.5.11/server/mvcc/watchable_store.go#L500-L504
+func (w *watcher) Progress(ctx context.Context) {
+	w.RLock()
```
marco6, line 255: While it might be sound in this case, taking an RLock and than writing to a field (posting to the progress channel) sounds like a bug IMHO. If we are reading, we should really not write.

pkg/kine/server/watch.go
```diff
@@ -4,49 +4,78 @@ import (
...
+	wg       sync.WaitGroup
+	backend  Backend
+	server   etcdserverpb.Watch_WatchServer
+	watches  map[int64]func()
+	progress map[int64]chan<- int64
 }
 
 func (w *watcher) Start(ctx context.Context, r *etcdserverpb.WatchCreateRequest) {
+	if r.WatchId != clientv3.AutoWatchID {
+		logrus.Warnf("WATCH START id=%d ignoring request with client-provided id", r.WatchId)
```
marco6, line 75: If we are ignoring the request, shouldn't we send and error message?

pkg/kine/server/types.go
```diff
@@ -39,3 +41,10 @@ type Event struct {
 	KV     *KeyValue
 	PrevKV *KeyValue
 }
+
+type WatchResult struct {
+	CurrentRevision int64
+	CompactRevision int64
+	Events          <-chan []*Event
+	Errorc          <-chan error
```
marco6, line 49: I think we should remove this field. See comment in the usage part.

pkg/kine/sqllog/sqllog.go
```diff
@@ -354,17 +363,24 @@ func (s *SQLLog) ttl(ctx context.Context) {
...
+func (s *SQLLog) Watch(ctx context.Context, key string, startRevision int64) (server.WatchResult, error) {
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Watch", otelName))
 	defer span.End()
 	span.SetAttributes(
 		attribute.String("key", key),
 		attribute.Int64("startRevision", startRevision),
 	)
 
+	res := make(chan []*server.Event, 100)
+	errc := make(chan error, 1)
```
marco6, line 375: This channel is used only in the synchronous part of this request (i.e. before starting the background job that actually watches).

So, it actually has no real advantage over just returning an error from the function given that once `SQLLog.Watch` returns, no code will ever notify this channel.

Also, it is never closed, which is in general a problem.

pkg/kine/sqllog/sqllog.go
```diff
@@ -71,6 +72,7 @@ type SQLLog struct {
 	d           Dialect
 	broadcaster broadcaster.Broadcaster[[]*server.Event]
 	notify      chan int64
+	currentRev  int64
```
marco6, line 75: This seems like a mistake. Caching something we do not have control over will result in weird errors along the way. 

For example, imagine that we reject a watch request asking for version `x`. Now what happens if `x` was compacted, but the polling interval didn't expire yet? We will either return a stale progress report or return an error with a version which sound valid but it isn't.

It also sounds like premature optimization: do we have data hinting that fetching the "real" revision would be expensive?

pkg/kine/server/watch.go
```diff
@@ -4,49 +4,78 @@ import (
...
-			w.Start(ws.Context(), msg.GetCreateRequest())
-		} else if msg.GetCancelRequest() != nil {
-			logrus.Debugf("WATCH CANCEL REQ id=%d", msg.GetCancelRequest().GetWatchId())
-			w.Cancel(msg.GetCancelRequest().WatchId, nil)
+		if cr := msg.GetCreateRequest(); cr != nil {
+			w.Start(ws.Context(), cr)
+		}
+		if cr := msg.GetCancelRequest(); cr != nil {
+			logrus.Tracef("WATCH CANCEL REQ id=%d", cr.WatchId)
+			w.Cancel(cr.WatchId, 0, 0, nil)
```
marco6, line 55: The two arguments to `Cancel` here are populated only once and are mostly `0`. This is a hint for me to say that they shouldn't be there in the first place. I think `Cancel` can fetch what it needs, if it needs.

pkg/kine/server/watch.go
```diff
@@ -4,49 +4,78 @@ import (
...
 
+// explicit interface check
+var _ etcdserverpb.WatchServer = (*KVServerBridge)(nil)
+
+// getProgressReportInterval returns the configured progress report interval, with some jitter
+func (s *KVServerBridge) getProgressReportInterval() time.Duration {
+	// add rand(1/10*notifyInterval) as jitter so that kine will not
+	// send progress notifications to watchers at the same time even when watchers
+	// are created at the same time.
+	jitter := time.Duration(rand.Int63n(int64(s.limited.notifyInterval) / 10))
```
marco6, line 29: Given that a polling starts when a watch arrives, do we really need a jitter? I think that requests are random enough not to care.

pkg/kine/server/watch.go
```diff
@@ -4,49 +4,78 @@ import (
...
-		backend: s.limited.backend,
-		watches: map[int64]func(){},
+		server:   ws,
+		backend:  s.limited.backend,
+		watches:  map[int64]func(){},
+		progress: map[int64]chan<- int64{},
 	}
 	defer w.Close()
 
+	go wait.PollInfiniteWithContext(ws.Context(), s.getProgressReportInterval(), w.ProgressIfSynced)
```
marco6, line 42: The method is deprecated, but in general we are adding a rather big dependency for a very simple functionality.

Can't we just write a polling loop here? I think if would also be cleaner.

pkg/kine/server/watch.go
```diff
@@ -4,49 +4,78 @@ import (
...
+	sync.RWMutex
 
-	wg      sync.WaitGroup
-	backend Backend
-	server  etcdserverpb.Watch_WatchServer
-	watches map[int64]func()
+	wg       sync.WaitGroup
+	backend  Backend
+	server   etcdserverpb.Watch_WatchServer
+	watches  map[int64]func()
```
marco6, line 69: Do we need 2 synchronized maps? Maybe we could add a type (even an inline one) and use a single one like:

```go
watches map[int64]struct{
  cancel func()
  progress chan<- int64
}
```

marco6 (changes requested):
Hi Louise! Thanks for the effort on this new requirement. First pass below:

pkg/kine/sqllog/sqllog.go
louiseschmidtgen: Our friends at kine [argued that](https://github.com/k3s-io/kine/pull/268):
```
Optimize CurrentRevision checks by sharing current revision marker from poll loop
Progress notifications query the current revision frequently; this can be optimized by exposing the poll loop's current revision once the loop has been started.
```
Our ProgressNotify interval is 5 seconds and there is no hard limit on the amount of watchers you can have.
If you think about deployments with 150,000 pods (number comes from [here](https://kubernetes.io/docs/setup/best-practices/cluster-large/)), would it still be cheap to get this read?

pkg/kine/server/watch.go
louiseschmidtgen: Yes, this is a good idea 👍🏼

pkg/kine/server/watch.go
louiseschmidtgen: I believe this could make sense for large deployments. If we create a deployment of an application with say 10,000 pods, watchers are started at the same time.

pkg/kine/server/watch.go
louiseschmidtgen: Looking at the upstream implementation there exists a RWlock on the progress report as well: https://github.com/etcd-io/etcd/blob/v3.5.11/server/mvcc/watchable_store.go#L54.

It is likely that we will have a high volume of progress reports, compared to small amount of watchers added/created for large deployments. In those cases the RWmutex would make sense to improve concurrency for readers.

pkg/kine/server/watch.go
louiseschmidtgen: Hm, yes it's a shortcut. The poll loop simplifies things nicely.

pkg/kine/server/watch.go
louiseschmidtgen: From discussion on MM, removing the jitter until we find evidence for its need.

pkg/kine/sqllog/sqllog.go
louiseschmidtgen: In MM agreed on removing caching.

pkg/kine/server/watch.go
louiseschmidtgen: I like the idea behind it, but in reality the nesting makes it quite inconvenient to iterate over in many places and doesn't help readability in the code.

marco6 (approved):
Looking way better, thanks Louise!

PR #216, opened 2025-01-09
Add database package

marco6:
This PR moves the `prepared` package and the implementation of prepared to a dedicated and more generic `database` package, so that it is possible to add other wrappers that add functionality on the way.

This was part of #180 and was split out so that the PR is smaller, and other functionality can use the right interface even if batching is not merged yet.

This PR does not change functionality and should not affect performance in any way.

Also, there is a small difference from the original version of this code in #180: `Wrap` method is no longer required for this to work, instead `NewPrepared` accepts a generic interface and acts as the wrapper itself. If this PR gets in this change will be reflected in #180 as well.


pkg/database/interface.go
```diff
@@ -0,0 +1,42 @@
+package database
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+)
+
+var errDBClosed = errors.New("sql: database is closed")
```
louiseschmidtgen, line 9: Where does this error get returned, does it get sent back to the api-server?

pkg/database/prepared.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	db.mu.Lock()
+	defer db.mu.Unlock()
+
+	if db.underlying == nil {
+		return nil, errDBClosed
+	}
+
+	// Check again if the query was prepared during locking
```
louiseschmidtgen, line 88: Why do we need to check this again?

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
+	"github.com/canonical/k8s-dqlite/pkg/database"
+)
+
+func TestPreparedPrepare(t *testing.T) {
```
louiseschmidtgen, line 11: Can we use a better name here? Maybe `TestPreparingStatements`?

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+	defer stmt1.Close()
+
+	stmt2, err := db.PrepareContext(ctx, expectedQuery)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer stmt2.Close()
+
+	if stmts := driver.stmts.Load(); stmts != 2 {
+		t.Errorf("invalid open statements: want 1, got %d", stmts)
```
louiseschmidtgen, line 34:
```suggestion
		t.Errorf("invalid open statements: want 2, got %d", stmts)
```
Should this be 2?

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer stmt1.Close()
+
+	stmt2, err := db.PrepareContext(ctx, expectedQuery)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer stmt2.Close()
```
louiseschmidtgen, line 31: What is the reason for testing two statements?

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if stmts := driver.stmts.Load(); stmts != 2 {
```
louiseschmidtgen, line 33: Can we be more precise and check that the prepared statement = expectedQuery string? Same goes for the other similar tests.

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+	defer db.Close()
+
+	rows, err := db.QueryContext(ctx, "query 1")
+	if err != nil {
+		t.Error(err)
+	}
+	rows.Close()
+
+	if stmts := driver.stmts.Load(); stmts != 1 {
+		t.Errorf("unexpected number of open statements: want %d, got %d", 1, stmts)
```
louiseschmidtgen, line 53:
```suggestion
		t.Errorf("unexpected number of open statements: want 1, got %d", stmts)
```
Let's use consistent format across our tests.

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	_, err = db.ExecContext(ctx, "query 2")
+	if err != nil {
+		t.Error(err)
+	}
+
+	if stmts := driver.stmts.Load(); stmts != 2 {
+		t.Errorf("unexpected number of open statements: want 2, got %d", stmts)
+	}
```
louiseschmidtgen, line 91: Why two queries here?

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		_, err = tx.ExecContext(ctx, "query 2")
+		if err != nil {
+			return (err)
+		}
+		return nil
+	}
+
+	for i := 0; i < 5; i++ {
```
louiseschmidtgen, line 121: Why are we using 5?

pkg/database/prepared_test.go
```diff
@@ -0,0 +1,139 @@
...
+		if err := transaction(); err != nil {
+			t.Error(err)
+		}
+	}
+
+	if stmts := driver.stmts.Load(); stmts != 4 {
+		t.Errorf("unexpected number of open statements: want 2, got %d", stmts)
```
louiseschmidtgen, line 128: 4 or 2?

louiseschmidtgen (commented):
Great work Marco! I love the direction this is going. This helps code readability greatly.
I've left some first questions and comments.

pkg/database/interface.go
marco6: This might if we have a race condition during shutdown. I really hope sooner or later to be able to confidently say that this does not happen, but in order to do that, we should make sure that the order of destroy is followed strictly like:
 - stop the gRPC server
 - stop the services around the db (TTL, Watchers, Compaction, etc...)
 - close the database driver
 - stop dqlite
 
 But right now I think it might happen that some of those phases overlap. Consider that this error is also fired by a normal `*sql.DB` instance and there isn't a proper constant to check (which means that the only way to catch this error is by checking the `Err()` string, in which case there is no way to distinguish the standard library one from this one).
 
This is however relevant only for tests as having few errors during shutdown means that the shutdown of the storage layer is happening before the shutdown of `kube-server` which would be quite weird (even if the shutdown is clean, how can the `kube-server` works if there is no storage layer?)

pkg/database/prepared_test.go
marco6: Well I could have used only one but it seemed like good practice to make sure that at least 2 are present? I don't know, testing one and many? I'm really not sure here.

pkg/database/prepared_test.go
marco6: lol copy paste madness ahahaha

pkg/database/prepared_test.go
marco6: Not sure what was the thinking but it should be at least 3 to check that statements are not prepared for each transaction, but reused.

pkg/database/prepared_test.go
marco6: I went for `TestPreparedDistinctStmts` to highlight better the purpose of the test and keep the `TestPreparedXXX` form.

pkg/database/driver_test.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (t *testConn) Close() error { return nil }
+
+type testStmt struct {
+	driver *testDriver
+	query  string
+}
+
+func (t *testStmt) NumInput() int { return 0 }
```
louiseschmidtgen, line 110: What is this function for?

pkg/database/interface.go
```diff
@@ -0,0 +1,42 @@
...
+
+type Interface interface {
+	QueryContext(ctx context.Context, query string, args ...any) (*sql.Rows, error)
+	ExecContext(ctx context.Context, query string, args ...any) (result sql.Result, err error)
+
```
louiseschmidtgen, line 14: Please remove the extra spaces in interface definitions.

pkg/database/interface.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+	"database/sql"
+	"errors"
+)
+
+var errDBClosed = errors.New("sql: database is closed")
+
+type Interface interface {
```
louiseschmidtgen, line 11: Optionally, you can add a little comment regarding the purpose of each interface.

pkg/database/prepared.go
```diff
@@ -0,0 +1,170 @@
...
+var otelTracer trace.Tracer
+
+func init() {
+	otelTracer = otel.Tracer(otelName)
+}
+
+type preparedDb[T Transaction] struct {
+	underlying Wrapped[T]
+	mu         sync.RWMutex
+	store      map[string]*sql.Stmt
```
louiseschmidtgen, line 25:
```suggestion
	cache      map[string]*sql.Stmt
```
Consider renaming store to cache?

pkg/database/prepared.go
louiseschmidtgen: Could you add that explanation in the comment please?

pkg/database/prepared.go
louiseschmidtgen, line 25: Would we ever need to remove any old/unused statements from cache?
We should be limited to the queries in generic and never have more than one of each cached right?
When we change a query in the future that will require a restart of the driver which means that we start with a new cache.

louiseschmidtgen (commented):
Overall LGTM, a few minor questions and comments:

pkg/database/driver_test.go
marco6: This is necessary for `testStmt` to implement the `Stmt` interface.

pkg/database/prepared.go
marco6: Yeah we don't have an "eviction" strategy for the very reason that we have few (`< 50`) queries and that we want all of them prepared. Some databases have a limit on prepared statements per connection (e.g. mysql), so if this was to become a "real" package, we would need to think about how to do that (LRU maybe?).

Right now I prefer to keep complexity low and postpone that.

pkg/database/prepared.go
marco6: I added some details, I hope it is better now.

pkg/database/interface.go
marco6: I can't really think about anything which is not implied by the name like: 

```
Interface represent the interface to a *sql.DB database
```

Which seems not very useful for a `database.Interface` interface? Do you have any suggestion?

louiseschmidtgen: Maybe `Interface that defines the methods for interacting with the underlying datastore` but that doesn't tell you much more than what you already see. Let's skip it then...

louiseschmidtgen (approved):
Great work Marco! Thanks for addressing my comments 🚀

PR #217, opened 2025-01-09
Remove useless field LeaderElect

marco6:
This PR removes `ETCDConfig.LeaderElect` field as it was never read (and thus useless).


louiseschmidtgen (approved):
LGTM, thank you for cleaning up the etcd leftovers.

PR #220, opened 2025-01-09
Remove TranslateErr from Generic

marco6:
This PR removes `TranslateErr` from `Generic` as that was a goofy way to return `ErrKeyExists` to the server.

We are now managing that by explicitly returning a bool from transactions. Also our queries do not fail any more, but they just don't insert anything (which is a requirement for batching).


louiseschmidtgen (approved):
Fantastic, we handle things much better now! 🚀

PR #222, opened 2025-01-16
refactor: remove `generic` package

marco6:
This PR removes the `generic` package and merges that with `sqlite`.

It also takes this opportunity to move many driver-unrelated stuff up in the instantiation chain, so that `sqlite` is now just a driver around a `*sql.DB` connection that is able to "talk" SQLite  sql dialect.

### DIFF
--- a/pkg/kine/drivers/dqlite/dqlite.go
+++ b/pkg/kine/drivers/dqlite/dqlite.go
@@ -24,16 +24,16 @@ func init() {
 
 type Driver = sqlite.Driver
 
-type DriverOptions struct {
+type DriverConfig struct {
 	DB      database.Interface
 	ErrCode func(error) string
 }
 
-func NewDriver(ctx context.Context, options *DriverOptions) (*Driver, error) {
+func NewDriver(ctx context.Context, config *DriverConfig) (*Driver, error) {
 	logrus.Printf("New driver for dqlite")
 
-	drv, err := sqlite.NewDriver(ctx, &sqlite.DriverOptions{
-		DB:         options.DB,
+	drv, err := sqlite.NewDriver(ctx, &sqlite.DriverConfig{
+		DB:         config.DB,
 		LockWrites: true,
 		Retry:      dqliteRetry,
 	})
@@ -41,7 +41,7 @@ func NewDriver(ctx context.Context, options *DriverOptions) (*Driver, error) {
 		return nil, err
 	}
 
-	if err := migrate(ctx, options.DB); err != nil {
+	if err := migrate(ctx, config.DB); err != nil {
 		return nil, errors.Wrap(err, "failed to migrate DB from sqlite")
 	}
 
--- a/pkg/kine/drivers/dqlite/dqlite.go
+++ b/pkg/kine/drivers/dqlite/dqlite.go
@@ -30,7 +30,7 @@ type DriverOptions struct {
 }
 
 func NewDriver(ctx context.Context, options *DriverOptions) (*Driver, error) {
-	logrus.Printf("New kine for dqlite")
+	logrus.Printf("New driver for dqlite")
 
 	drv, err := sqlite.NewDriver(ctx, &sqlite.DriverOptions{
 		DB:         options.DB,
--- a/pkg/kine/drivers/dqlite/dqlite.go
+++ b/pkg/kine/drivers/dqlite/dqlite.go
@@ -8,8 +8,8 @@ import (
 
 	"github.com/canonical/go-dqlite/v2"
 	"github.com/canonical/go-dqlite/v2/driver"
+	"github.com/canonical/k8s-dqlite/pkg/database"
 	"github.com/canonical/k8s-dqlite/pkg/kine/drivers/sqlite"
-	"github.com/canonical/k8s-dqlite/pkg/kine/sqllog"
 	"github.com/mattn/go-sqlite3"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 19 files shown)


marco6:
Depends on #218

marco6:
Ready for review as #218 was merged.

pkg/kine/drivers/dqlite/dqlite.go
```diff
@@ -95,50 +99,65 @@ func migrate(ctx context.Context, newDB *sql.Conn) (exitErr error) {
 	}
 	defer oldDB.Close()
 
-	oldData, err := oldDB.QueryContext(ctx, "SELECT id, name, created, deleted, create_revision, prev_revision, lease, value, old_value FROM kine")
+	oldData, err := oldDB.QueryContext(ctx, `
+SELECT id, name, created, deleted, create_revision, prev_revision, lease, value, old_value
+FROM kine
+ORDER BY id ASC`)
```
louiseschmidtgen, line 105: Could we ever reach a limit of how many rows this query can return at once? Would we need a paginated request?

pkg/kine/drivers/sqlite/sqlite.go
```diff
@@ -6,28 +6,35 @@ import (
...
 
-	"github.com/canonical/k8s-dqlite/pkg/kine/drivers/generic"
+	"github.com/canonical/k8s-dqlite/pkg/database"
 	"github.com/mattn/go-sqlite3"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
-func NewDriver(ctx context.Context, driverName, dataSourceName string, connectionPoolConfig *generic.ConnectionPoolConfig) (*generic.Generic, error) {
+func NewDriver(ctx context.Context, options *DriverOptions) (*Driver, error) {
```
louiseschmidtgen, line 15: Should we call this the `DriverConfig`? Things like the `ErrCode` or the `DB` don't come from the rootCmd options.

pkg/server/server.go
```diff
@@ -27,8 +29,9 @@ type Server struct {
 
 	backend server.Backend
 
-	// kineConfig is the configuration to use for starting kine against the dqlite application.
-	kineConfig endpoint.Config
+	// serverConfig is the configuration to use for starting kine against the dqlite application.
```
louiseschmidtgen, line 32: We should update the comments now to be more general and remove dqlite.

pkg/server/server.go
```diff
@@ -369,20 +406,118 @@ func (s *Server) Start(ctx context.Context) error {
 	}
 	logrus.WithFields(logrus.Fields{"id": s.app.ID(), "address": s.app.Address()}).Print("Started dqlite")
 
-	logrus.WithField("config", s.kineConfig).Debug("Starting kine")
-	_, backend, err := endpoint.ListenAndReturnBackend(ctx, s.kineConfig)
+	logrus.WithField("config", s.serverConfig).Debug("Starting kine")
```
louiseschmidtgen, line 409:
```suggestion
	logrus.WithField("config", s.serverConfig).Debug("Starting k8s-dqlite server")
```

pkg/server/server.go
```diff
@@ -369,20 +406,118 @@ func (s *Server) Start(ctx context.Context) error {
...
+		ListenAddress: s.serverConfig.ListenAddress,
+		Server:        server.New(backend, s.serverConfig.NotifyInterval),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to start kine: %w", err)
+	}
+
 	go s.watchAvailableStorageSize(ctx)
 
+	logrus.WithField("address", s.serverConfig.ListenAddress).Print("Started kine")
```
louiseschmidtgen, line 446:
```suggestion
	logrus.WithField("address", s.serverConfig.ListenAddress).Print("Started k8s-dqlite server")
```

pkg/server/server.go
```diff
@@ -369,20 +406,118 @@ func (s *Server) Start(ctx context.Context) error {
...
+	if err := db.PingContext(ctx); err != nil {
+		logrus.Errorf("failed to ping connection: %v", err)
+		db.Close()
+		return nil, err
+	}
+
+	return db, nil
+}
+
+// func openKineStorageBackend(ctx context.Context, config *Config) (server.Backend, error) {
```
louiseschmidtgen, line 490: Can we remove this section please?

test/util_test.go
```diff
@@ -108,59 +94,71 @@ func newKineServer(ctx context.Context, tb testing.TB, options *kineOptions) *ki
...
 		client:         client,
 		backend:        backend,
 		dqliteListener: dqliteListener,
 	}
 }
 
-func startSqlite(_ context.Context, tb testing.TB, dir string) (*endpoint.Config, *sql.DB) {
-	dbPath := path.Join(dir, "data.db")
+func startSqlite(ctx context.Context, tb testing.TB, dir string) (*sqlite.Driver, *sql.DB) {
+	dbPath := path.Join(dir, "data.db?_journal=WAL&_synchronous=FULL&_foreign_keys=1")
```
louiseschmidtgen, line 144: This reverts your last PR, do we need to rebase?

test/util_test.go
```diff
@@ -180,11 +178,18 @@ func startDqlite(ctx context.Context, tb testing.TB, dir string, listener *instr
 		tb.Fatal(err)
 	}
 
-	return &endpoint.Config{
-		Listener:       fmt.Sprintf("unix://%s/kine.sock", dir),
-		Endpoint:       fmt.Sprintf("dqlite://k8s?driver-name=%s", app.Driver()),
-		NotifyInterval: 5 * time.Second,
-	}, db
+	if err := db.PingContext(ctx); err != nil {
```
louiseschmidtgen, line 181: can we use the open and test function that does the ping for the test?

pkg/server/server.go
```diff
@@ -27,8 +29,9 @@ type Server struct {
 
 	backend server.Backend
 
-	// kineConfig is the configuration to use for starting kine against the dqlite application.
-	kineConfig endpoint.Config
+	// serverConfig is the configuration to use for starting kine against the dqlite application.
+	serverConfig         *ServerConfig
+	connectionPoolConfig *ConnectionPoolConfig
```
louiseschmidtgen, line 34: Would it make sense to collapse some of these configuration options into the `serverConfig` as well? For example the `connectionPoolConfig`?

louiseschmidtgen (commented):
Nice work Marco! Left some comments:

pkg/kine/drivers/sqlite/sqlite.go
marco6: I've had a discussion in the Go MM channel and while these two concepts are kind of overlapping, it seems like it's best to have Config here. 

I will also issue a separate PR to fix this naming issue in the rest of the codebase.

pkg/server/server.go
marco6: I'll remove this comment as it doesn't really add much to the context.

pkg/server/server.go
marco6: I think we can do that in a next phase as this PR is already big enough.

test/util_test.go
marco6: Ah it's already rebased, I must have missed something during merge :-( Nice catch!

test/util_test.go
marco6: That is unexported, so we cannot. Some copying is fine in this case.

louiseschmidtgen (approved):
LGTM, thank you Marco!

Issue #196, opened 2024-10-30
Excessive memory usage by k8s-dqlite


sbidoul:
Hello,

Since a few weeks (sorry I cannot be more precise), I notice an increased memory utilization over time on dqlite nodes.

Memory usage on the node over a week looks like this, and I can attribute the increase in RAM use to the k8s-dqlite process.

![image](https://github.com/user-attachments/assets/33cbc28d-2e26-421c-9d01-9e52d7965d93)

The cluster has been on 1.29.x since early 2024, but the leak only started to manifest itself recently, presumably following an automatic minor snap update.

Is there anything I can do to help diagnosing this?

jcjveraa:
Hi there,

Same here, on my 3 node experimental cluster with 16GB ram per node, while running barely any workloads the memory usage for dqlite is 6.9% of total memory - 1.1 GB for a 12MB database?

Commands used (note: by the time I ran this it was 7.0% - 4th column of the output of `ps aux` ).

``` 
$ ps aux | grep dqlite
root         818  5.5  7.0 2406272 1147972 ?     Ssl  Nov09 538:51 /snap/microk8s/7394/bin/k8s-dqlite --storage-dir=/var/snap/microk8s/7394/var/kubernetes/backend/ --listen=unix:///var/snap/microk8s/7394/var/kubernetes/backend/kine.sock:12379

$ microk8s dbctl backup 
$ tar -xvzf backup-2024-11-16-13-44-42.tar.gz
$ cd backup-2024-11-16-13-44-42
$ du -h
12M	.
```

sbidoul:
@jcjveraa to be clear, my issue is not about high memory usage, but about an continuous memory usage increase, suggesting a memory leak.

jcjveraa:
> @jcjveraa to be clear, my issue is not about high memory usage, but about an continuous memory usage increase, suggesting a memory leak.

Yes same here, in the meantime I’ve switched to k3s in high availability (etdc) mode, with the same workload, and memory consumption is completely flat. For me “my workloads + microk8s” was certainly the cause for the memory leak, and I suspect it to be in dqlite.

svetlak0f:
Yes, I noticed the same behavior for both of my microk8s clusters
![image](https://github.com/user-attachments/assets/063199b1-14c6-4a40-837d-b5bf7f88e22c)

When inspecting with htop, there was a huge RAM consumption by k8s-dqlite, which disappeared after restarting the DB daemon, but returned after some time:
```
[...]

PR #230, opened 2025-01-29
Default and minimum snapshot parameters

louiseschmidtgen:
## Default and minimum snapshot parameters

For context see issue: https://github.com/canonical/k8s-dqlite/issues/196

The currently used Dqlite snapshot parameters are set too high for small clusters.
This PR adjusts the default snapshot parameter's used.
It also implements some guardrails for setting the parameters via the tuning config.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -61,7 +61,6 @@ const (
 	minThreshold     = 384
 	defaultTrailing  = 1024
 	minTrailing      = 512
-	scalingFactor    = 2
 )
 
 // New creates a new instance of Server based on configuration.
@@ -251,18 +250,25 @@ func New(
 		if v := tuning.Snapshot; v != nil {
 			logrus.WithFields(logrus.Fields{"threshold": v.Threshold, "trailing": v.Trailing}).Print("Initial dqlite raft snapshot parameters before adjustments")
 
-			if v.Trailing < minTrailing && v.Threshold > minThreshold {
-				snapshotParameters.Trailing = v.Threshold * scalingFactor
-				snapshotParameters.Threshold = v.Threshold
-			} else if v.Threshold < minThreshold && v.Trailing > minTrailing {
-				snapshotParameters.Threshold = v.Trailing * 1 / scalingFactor
-				snapshotParameters.Trailing = v.Trailing
-			} else if v.Trailing < minTrailing && v.Threshold < minThreshold {
-				snapshotParameters.Threshold = minThreshold
+			snapshotParameters.Threshold = v.Threshold
+			snapshotParameters.Trailing = v.Trailing
+
+			if v.Trailing < minTrailing {
 				snapshotParameters.Trailing = minTrailing
-			} else {
-				snapshotParameters.Threshold = v.Threshold
-				snapshotParameters.Trailing = v.Trailing
+				logrus.WithFields(logrus.Fields{"adjustedTrailing": snapshotParameters.Trailing}).Warning("Trailing value is too low, setting to minimum value")
+			}
+			if v.Threshold == 0 {
+				snapshotParameters.Threshold = v.Trailing / 2
+				logrus.WithFields(logrus.Fields{"adjustedThreshold": snapshotParameters.Threshold}).Warning("Threshold value is zero, setting to half of trailing value")
+
+			} else if v.Threshold < minThreshold {
+				snapshotParameters.Threshold = minThreshold
+				logrus.WithFields(logrus.Fields{"adjustedThreshold": snapshotParameters.Threshold}).Warning("Threshold value is too low, setting to minimum value")
+
+			}
+			if v.Threshold > v.Trailing {
+				snapshotParameters.Threshold = v.Trailing
+				logrus.WithFields(logrus.Fields{"adjustedThreshold": snapshotParameters.Threshold}).Warning("Threshold value is higher than trailing value, setting to trailing value")
 			}
 		}
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -245,11 +245,11 @@ func New(
 		if v := tuning.Snapshot; v != nil {
 			logrus.WithFields(logrus.Fields{"threshold": v.Threshold, "trailing": v.Trailing}).Print("Configure dqlite raft snapshot parameters")
 			if v.Trailing < v.Threshold {
-				return nil, fmt.Errorf("trailing snapshot threshold must be less than snapshot threshold")
+				return nil, fmt.Errorf("the snapshot's trailing parameter must be less than the threshold parameter")
 			}
 			// ensure that the threshold is at least 384 and trailing is at least 512 to avoid CPU utilization issues.
 			if v.Threshold < min_trailing || v.Trailing < min_trailing {
-				logrus.WithFields(logrus.Fields{"threshold": min_threshold, "trailing": min_trailing}).Print("Adjusting dqlite raft snapshot parameters")
+				logrus.WithFields(logrus.Fields{"threshold": min_threshold, "trailing": min_trailing}).Print("Using dqlite raft minimum snapshot parameters, as provided value(s) are too low")
 				v.Threshold = min_threshold
 				v.Trailing = min_trailing
 			}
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -56,6 +56,13 @@ var expectedFilesDuringInitialization = map[string]struct{}{
 	"tuning.yaml":    {},
 }
 
+const (
```
marco6, line 59: Can we not use snake_case in Go? Let's be uniform with the rest and use (lowercase)CamelCase for (private)constants.

pkg/server/server.go
```diff
@@ -237,6 +244,15 @@ func New(
 
 		if v := tuning.Snapshot; v != nil {
 			logrus.WithFields(logrus.Fields{"threshold": v.Threshold, "trailing": v.Trailing}).Print("Configure dqlite raft snapshot parameters")
+			if v.Trailing < v.Threshold {
```
marco6, line 247: I think this should come after the sanitization below. This helps catching the right behaviour.

I also think we should not error (and completely shut down the node) but instead log a warning and compute a suitable value instead. A rule of thumb might be `Threshold = Trailing * 3 / 4`. Also `Threshold = Trailing / 2` could work.

pkg/server/server.go
```diff
@@ -237,6 +244,15 @@ func New(
...
+				return nil, fmt.Errorf("the snapshot's trailing parameter must be less than the threshold parameter")
+			}
+			// ensure that the threshold is at least 384 and trailing is at least 512 to avoid CPU utilization issues.
```
marco6, line 250: I think the code is clear enough not to need any comment.

pkg/server/server.go
```diff
@@ -237,6 +244,15 @@ func New(
...
+			if v.Threshold < min_trailing || v.Trailing < min_trailing {
```
marco6, line 251: I think the two conditions should be separated. Otherwise we would overwrite the requested behaviour without reason. For example imagine something like:

```yaml
threshold : 200
trailing: 1024
```

This logic would use `trailing = 512` even if the user-provided value is ok.

pkg/server/server.go
```diff
@@ -251,6 +267,11 @@ func New(
 		// these are set in the kine endpoint config below
 		compactInterval = tuning.KineCompactInterval
 		pollInterval = tuning.KinePollInterval
+	} else {
```
marco6, line 270: Can we re-organize the code in such a way that we don't need to write `append` twice? Something like:

```go
snapshotParameters := dqlite.SnapshotParams{ // declare default
	Threshold: default_threshold,
	Trailing:  default_trailing,
}
if exists, err := fileExists(dir, "tuning.yaml"); err != nil { // logic to load user decisions
... all the logic above here ....
}
options = append(options, app.WithSnapshotParams(dqlite.SnapshotParams{
	Threshold: v.Threshold,
	Trailing:  v.Trailing,
})
...
```

pkg/server/server.go
```diff
@@ -56,6 +56,13 @@ var expectedFilesDuringInitialization = map[string]struct{}{
 	"tuning.yaml":    {},
 }
 
+const (
+	default_threshold = 512
+	min_threshold     = 384
+	default_trailing  = 1024
+	min_trailing      = 512
```
marco6, line 63: I'm not convinced in general that the best value for threshold is half trailing. Could we test using a slightly higher value like `3 /4` of it (in this case `738`)? I think this will reduce the amount of snapshots required without harming too much the protocol.

marco6 (changes requested):
Thanks Louise, this seems like a good direction. Few thoughts below.

pkg/server/server.go
louiseschmidtgen: Okay, I'll test it and adjust the factor!

louiseschmidtgen: As discussed in MM the results for 3/4 aren't better than 1/2. We'll stick with this factor.

pkg/server/server.go
```diff
@@ -236,11 +249,21 @@ func New(
...
-			options = append(options, app.WithSnapshotParams(dqlite.SnapshotParams{
-				Threshold: v.Threshold,
-				Trailing:  v.Trailing,
-			}))
+			logrus.WithFields(logrus.Fields{"threshold": v.Threshold, "trailing": v.Trailing}).Print("Initial dqlite raft snapshot parameters before adjustments")
+
+			if v.Trailing < minTrailing && v.Threshold > minThreshold {
+				snapshotParameters.Trailing = v.Threshold * scalingFactor
+				snapshotParameters.Threshold = v.Threshold
+			} else if v.Threshold < minThreshold && v.Trailing > minThreshold {
```
HomayoonAlimohammadi, line 257: I _think_ this should be:
```suggestion
			} else if v.Threshold < minThreshold && v.Trailing > minTrailing {
```

pkg/server/server.go
```diff
@@ -236,11 +249,21 @@ func New(
...
 		if v := tuning.Snapshot; v != nil {
-			logrus.WithFields(logrus.Fields{"threshold": v.Threshold, "trailing": v.Trailing}).Print("Configure dqlite raft snapshot parameters")
-			options = append(options, app.WithSnapshotParams(dqlite.SnapshotParams{
-				Threshold: v.Threshold,
-				Trailing:  v.Trailing,
-			}))
+			logrus.WithFields(logrus.Fields{"threshold": v.Threshold, "trailing": v.Trailing}).Print("Initial dqlite raft snapshot parameters before adjustments")
+
+			if v.Trailing < minTrailing && v.Threshold > minThreshold {
+				snapshotParameters.Trailing = v.Threshold * scalingFactor
```
HomayoonAlimohammadi, line 255: Should this also change to `v.Trailing`?
```suggestion
				snapshotParameters.Trailing = v.Trailing * scalingFactor
```

pkg/server/server.go
```diff
@@ -236,11 +249,21 @@ func New(
...
+				snapshotParameters.Threshold = v.Threshold
+			} else if v.Threshold < minThreshold && v.Trailing > minTrailing {
+				snapshotParameters.Threshold = v.Trailing * 1 / scalingFactor
+				snapshotParameters.Trailing = v.Trailing
+			} else if v.Trailing < minTrailing && v.Threshold < minThreshold {
+				snapshotParameters.Threshold = minThreshold
+				snapshotParameters.Trailing = minTrailing
+			} else {
+				snapshotParameters.Threshold = v.Threshold
```
HomayoonAlimohammadi, line 264: Just as a question, I noticed that we set `Threshold` and `Trailing` to `minThreshold * scalingFactor` and `minTrailing * scalingFactor` _**only if one of them**_ is less than it's minimum. But when both are less than minimum we set both to the minimum value. Why is this?

pkg/server/server.go
```diff
@@ -236,11 +249,21 @@ func New(
...
-				Threshold: v.Threshold,
-				Trailing:  v.Trailing,
-			}))
+			logrus.WithFields(logrus.Fields{"threshold": v.Threshold, "trailing": v.Trailing}).Print("Initial dqlite raft snapshot parameters before adjustments")
+
+			if v.Trailing < minTrailing && v.Threshold > minThreshold {
+				snapshotParameters.Trailing = v.Threshold * scalingFactor
+				snapshotParameters.Threshold = v.Threshold
+			} else if v.Threshold < minThreshold && v.Trailing > minTrailing {
+				snapshotParameters.Threshold = v.Trailing * 1 / scalingFactor
```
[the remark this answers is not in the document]
louiseschmidtgen: This one is fine :)

pkg/server/server.go
louiseschmidtgen: The relationship for trailing:threshold is 1:2, so if one of the given parameters that the user gives us is to small we'll compute it from the other value.

pkg/server/server.go
louiseschmidtgen: Basically if one value is fine we compute a reasonable value for the other value in the first two statements, if both are too small we assign it the minimum values, if the user's values are fine we assign them, if we have no input we pass in default values.

Does that make sense?
Would you prefer me leaving a comment in the code or is it clear enough?

HomayoonAlimohammadi: Got it, thanks Louise! I believe the code is clear enough, no need for a comment here.

marco6 (approved):
Thanks Louise! LGTM

PR #235, opened 2025-02-06
allow compactions only on leader nodes

petrutlucian94:
We'll only allow compactions to be performed on dqlite leader nodes, checking if the local node is the leader whenever a compaction is requested.

*Also verify you have:*

### DIFF
--- a/pkg/kine/drivers/dqlite/driver.go
+++ b/pkg/kine/drivers/dqlite/driver.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"time"
 
 	"github.com/canonical/go-dqlite/v2"
 	"github.com/canonical/go-dqlite/v2/app"
@@ -36,18 +35,14 @@ func (d *Driver) Compact(ctx context.Context, revision int64) (err error) {
 	//
 	// WARNING: overridden methods will not be visible from the embedded
 	// (sqlite) driver.
-	start := time.Now()
 	isLeader, err := isLocalNodeLeader(ctx, d.goDqliteApp)
-	logrus.Printf(">>> compact isLocalNodeLeader duration : %v", time.Now().Sub(start))
 	if err != nil {
 		logrus.WithError(err).Warning("Couldn't determine whether the local node is the leader, allowing the compaction to proceed")
 	} else if !isLeader {
 		logrus.Info("skipping compaction on follower node")
 		return nil
 	}
-	err = d.Driver.Compact(ctx, revision)
-	logrus.Printf(">>> compact total duration : %v", time.Now().Sub(start))
-	return err
+	return d.Driver.Compact(ctx, revision)
 }
 
 type DriverConfig struct {
--- a/pkg/kine/drivers/dqlite/driver.go
+++ b/pkg/kine/drivers/dqlite/driver.go
@@ -37,7 +37,7 @@ func (d *Driver) Compact(ctx context.Context, revision int64) (err error) {
 	// (sqlite) driver.
 	isLeader, err := isLocalNodeLeader(ctx, d.goDqliteApp)
 	if err != nil {
-		logrus.WithError(err).Warning("couldn't determine if the local node has leader role, allowing the compaction to proceed")
+		logrus.WithError(err).Warning("Couldn't determine whether the local node is the leader, allowing the compaction to proceed")
 	} else if !isLeader {
 		logrus.Info("skipping compaction on follower node")
 		return nil
--- a/pkg/kine/drivers/dqlite/driver.go
+++ b/pkg/kine/drivers/dqlite/driver.go
@@ -3,10 +3,12 @@ package dqlite
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"os"
 	"strings"
 
 	"github.com/canonical/go-dqlite/v2"
+	"github.com/canonical/go-dqlite/v2/app"
 	"github.com/canonical/go-dqlite/v2/driver"
 	"github.com/canonical/k8s-dqlite/pkg/database"
 	"github.com/canonical/k8s-dqlite/pkg/kine/drivers/sqlite"
@@ -29,14 +31,18 @@ type DriverConfig struct {
 	ErrCode func(error) string
 }
 
-func NewDriver(ctx context.Context, config *DriverConfig) (*Driver, error) {
+func NewDriver(ctx context.Context, config *DriverConfig, goDqliteApp *app.App) (*Driver, error) {
 	logrus.Printf("New driver for dqlite")
 
 	drv, err := sqlite.NewDriver(ctx, &sqlite.DriverConfig{
 		DB:         config.DB,
 		LockWrites: true,
 		Retry:      dqliteRetry,
+		CompactAllowed: func(ctx context.Context) (bool, error) {
+			return compactAllowed(ctx, goDqliteApp)
+		},
 	})
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/server/server.go
```diff
@@ -309,6 +309,7 @@ func New(
 
 	kineConfig.Listener = listen
 	kineConfig.Endpoint = fmt.Sprintf("dqlite://k8s?%s", params.Encode())
+	kineConfig.GoDqliteApp = app
```
louiseschmidtgen, line 312: IMHO the app is not really a kineConfig.

pkg/kine/drivers/generic/generic.go
```diff
@@ -525,6 +527,17 @@ func (d *Generic) Delete(ctx context.Context, key string, revision int64) (rev i
...
+	if d.CompactAllowed != nil {
+		compactAllowed, err := d.CompactAllowed(ctx)
+		if err != nil {
+			return fmt.Errorf("compaction pre-check failed: %w", err)
+		}
+		if !compactAllowed {
+			// TODO: should we log something or return an error?
+			return nil
+		}
+	}
```
louiseschmidtgen, line 539: Let's call the `isLocalNodeLeader` function here.
If the check fails/ we're the leader >> we continue with compaction.
If we're not the leader you can log a debug message `Skipping compaction on follower node` and return a nil error.

louiseschmidtgen (commented):
Thanks Lucian! Left some first comments. 
Also please re-base once https://github.com/canonical/k8s-dqlite/pull/218 is merged.

pkg/kine/drivers/generic/generic.go
```diff
@@ -525,6 +527,17 @@ func (d *Generic) Delete(ctx context.Context, key string, revision int64) (rev i
 // After the call, any request for a version older than the given revision will return
 // a compacted error.
 func (d *Generic) Compact(ctx context.Context, revision int64) (err error) {
+	if d.CompactAllowed != nil {
```
louiseschmidtgen, line 530: We won't be able to call the go-dqlite app if we are using an sqlite driver.

pkg/server/server.go
petrutlucian94: Makes sense, we can pass it separately.

pkg/kine/drivers/generic/generic.go
petrutlucian94: Moving ``isLocalNodeLeader`` here seemed wrong since the ``generic`` driver is dqlite agnostic, which is why there's a separate dqlite driver.

petrutlucian94: Thanks for mentioning the failing unit test, I'll take care of it asap.

louiseschmidtgen: We're about to remove the generic driver, that's why :D https://github.com/canonical/k8s-dqlite/pull/222

louiseschmidtgen:
One more thing, could you please adjust the performance test to repeat a few times (maybe about 15 times)? You can just repeat the kube-burner part and keep using the same capture metrics call for all.
Currently, it runs for about two minutes. Compaction runs every 5 minutes. 
Once that's done in the artifact from the summary of the performance run we can have a look at the CPU usage from our three nodes test. That will tell us how much skipping the compaction on followers helps us.

pkg/kine/drivers/generic/generic.go
petrutlucian94: Oh, got it :)

petrutlucian94: > Let's call the isLocalNodeLeader function in here.

The ``Compact`` method is now part of the sqlite driver, we still cannot use dqlite specific logic there.

https://github.com/canonical/k8s-dqlite/blob/cceeb47eafee3669ae3b17b00f32fbd304344ded/pkg/kine/drivers/sqlite/driver.go#L573-L602

petrutlucian94: I've added ``IsLeader`` to the driver interface since we'll need that for the ``ttl`` check as well.

test/performance/tests/test_util/metrics.py
```diff
@@ -78,17 +81,32 @@ def configure_kube_burner(instance: harness.Instance):
     )
 
 
-def run_kube_burner(instance: harness.Instance):
+def run_kube_burner(instance: harness.Instance, iterations: int = 15):
```
louiseschmidtgen, line 84: Could we make the iterations a configurable parameter please?

.github/workflows/performance.yaml
```diff
@@ -114,12 +114,12 @@ jobs:
           name: performance-results
           path: ${{ github.workspace }}/test/performance/results
       - name: Prepare inspection reports
-        if: failure()
+        if: always()
```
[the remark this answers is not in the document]
petrutlucian94: k8s-dqlite logs

pkg/kine/drivers/dqlite/driver.go
```diff
@@ -29,14 +31,18 @@ type DriverConfig struct {
...
 
-func NewDriver(ctx context.Context, config *DriverConfig) (*Driver, error) {
+func NewDriver(ctx context.Context, config *DriverConfig, goDqliteApp *app.App) (*Driver, error) {
 	logrus.Printf("New driver for dqlite")
 
 	drv, err := sqlite.NewDriver(ctx, &sqlite.DriverConfig{
 		DB:         config.DB,
 		LockWrites: true,
 		Retry:      dqliteRetry,
+		IsLeader: func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
petrutlucian94: indeed, but it has to be part of the driver interface so that ``sqqlog`` can call it. The actual implementation is passed through the ``DriverConfig``, which is the only way of having specialized driver methods (just like the Retry callback).

For sqlite this will be a no-op, always assuming that the node is a leader.

.github/workflows/performance.yaml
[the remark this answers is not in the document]
louiseschmidtgen: Cool, it might give us the chance to look in the logs if we see unexpected results in the performance stats.

pkg/kine/drivers/dqlite/driver.go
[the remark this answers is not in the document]
louiseschmidtgen: Can we override the Compact() function in the Dqlite driver?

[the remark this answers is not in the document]
petrutlucian94: > Can we override the Compact() function in the Dqlite driver?

Sure, done. The only downside with how function overriding works in golang is that other ``sqlite.Driver`` methods will not be able to see the `dqlite.Driver`` ``Compact`` method. It will only be visible to external callers.

This is ok atm since ``Compact`` is not used by other driver methods but it might affect us in the future with other overriden methods.

pkg/kine/drivers/dqlite/driver.go
```diff
@@ -22,21 +24,41 @@ func init() {
...
+}
+
+func (d *Driver) Compact(ctx context.Context, revision int64) (err error) {
+	// Skip the compaction if we're not the leader.
+	//
+	// WARNING: overridden methods will not be visible from the embedded
+	// (sqlite) driver.
+	isLeader, err := isLocalNodeLeader(ctx, d.goDqliteApp)
+	if err != nil {
+		logrus.WithError(err).Warning("couldn't determine if the local node has leader role, allowing the compaction to proceed")
```
louiseschmidtgen, line 40:
```suggestion
		logrus.WithError(err).Warning("Couldn't determine whether the local node is the leader, allowing the compaction to proceed")
```
-> We have to capitalize to stay consistent with the other warnings in this repo.

louiseschmidtgen (commented):
Much better, thanks! Let's discuss performance metrics before we merge.

louiseschmidtgen:
Let's rebase on https://github.com/canonical/k8s-dqlite/pull/249 once it's ready.

test/util_test.go
```diff
@@ -193,7 +193,7 @@ func startDqlite(ctx context.Context, tb testing.TB, dir string, listener *instr
 
 	driver, err := dqlite.NewDriver(ctx, &dqlite.DriverConfig{
 		DB: database.NewPrepared(db),
-	})
+	}, nil)
```
marco6, line 196: Why passing nil here? I think we have the `app` so we should pass it down.

pkg/kine/drivers/dqlite/driver.go
```diff
@@ -22,14 +25,37 @@ func init() {
...
+		logrus.Info("skipping compaction on follower node")
+		return nil
+	}
+	err = d.Driver.Compact(ctx, revision)
+	logrus.Printf(">>> compact total duration : %v", time.Now().Sub(start))
+	return err
+}
 
 type DriverConfig struct {
 	DB      database.Interface
```
marco6, line 54: I have been wondering in the previous PR whose responsibility was to open the `*sql.DB` interface (i.e. call `sql.Open`).
The answer wasn't clear, so I went for the usual dependency inversion thing. However the addition of an `app.App` field in this struct seems like a good reason to lean in the other direction and call it inside of this method, so that only the app is passed down. 

It also kind of makes sense as the wrapping of a `sql.DB` depends on the system at hand (as the driver is the one who knows how to gain the best performance out of the db, not the caller).

We can do that either here or in another separate PR.

marco6 (changes requested):
Hi @petrutlucian94, this seems like a good direction. Few things to consider below

test/util_test.go
petrutlucian94: Oh, right. missed it.

pkg/kine/drivers/dqlite/driver.go
petrutlucian94: I see. ``ConnectionPoolConfig`` would also have to be moved to the dqlite driver, I think.

https://github.com/canonical/k8s-dqlite/blob/26442adc4de6e898ac1c3b34a1ce68817891fa67/pkg/server/server.go#L62-L91

I'd go with a separate PR.

marco6: Right. Let's do that later on.

marco6 (approved):
Thanks for the changes, LGTM!

PR #238, opened 2025-02-10
refactor: use `XXXConfig` where appropriate

marco6:
As a follow-up to #222, this PR ensures that naming of configuration is consistent. The idea behind the naming choice is:

 - use `XXXConfig` when at least one of the values passed is mandatory (and as such the struct is not and `Option`)
 - use `XXXOptions` when all the values are optional. In this case it should be possible to pass `nil` as well.
 
 It also takes the chance to rename and merge some files in the driver which were a leftover from a previous change.


louiseschmidtgen (approved):
LGTM

PR #239, opened 2025-02-10
Collect pprof cpu time data

petrutlucian94:
This PR will collect pprof profiling data from the performance tests, telling us how much time was spent in various k8s-dqlite functions. This allows us to identify bottlenecks and measure the cost of certain operations.

Note that k8s-dqlite already exposes a [net/http/pprof](https://pkg.go.dev/net/http/pprof) endpoint. The problem is that it collects data for a predetermined duration and we can't know for sure how long our perf tests are going to last. For this reason, we're adding the ``--profiling-dir`` flag. If set, we'll output the profiling data to that directory when the k8s-dqlite process stops, thus having data for the entire duration of the perf tests.

For now, we'll only retrieve the cpu times. In the future we may add other profiling data such as memory statistics.

To collect samples, pprof sends 100 ``SIGPROF`` signals per second and inspects the stack traces. It may not catch functions that return faster than that. The sample frequency may be adjusted using ``SetCPUProfileRate``, if needed we can make it configurable (that would imply an overhead and most probably won't be required).

*Also verify you have:*


test/performance/tests/test_util/config.py
```diff
@@ -43,6 +43,9 @@
 # Global kube-burner invocation timeout.
 KUBE_BURNER_TIMEOUT = os.getenv("TEST_KUBE_BURNER_TIMEOUT") or "10m"
 
+# How long to collect pprof profile data (seconds).
```
petrutlucian94, line 46: TODO: remove leftover.

test/performance/tests/test_util/metrics.py
```diff
@@ -39,6 +39,7 @@ def collect_metrics(instances: List[harness.Instance]):
             background=True,
         )
         process_dict[instance.id] = subprocess
+
```
petrutlucian94, line 42: TODO: leftover unrelated change.

petrutlucian94:
TODO:

* see why some ``/root/cpu_profile.raw`` files are missing
* ``Duration: 190.20s, Total samples = 24.49s (12.88%)``
  * it looks like only 25s out of 190s was sampled

petrutlucian94:
The ``base code snap`` perf test is expected to fail since it doesn't contain the new pprof setting.

petrutlucian94:
``pprof`` does have an impact on the profiled process, so we can also make this optional.

test/performance/tests/conftest.py
```diff
@@ -173,13 +173,13 @@ def instances(
...
-            )
-        else:
-            first_node.exec(["k8s", "bootstrap"])
+        if not bootstrap_config:
+            bootstrap_config = (config.MANIFESTS_DIR / "bootstrap-all.yaml").read_text()
+
+        first_node.exec(
+            ["k8s", "bootstrap", "--file", "-"],
+            input=str.encode(bootstrap_config),
+        )
```
louiseschmidtgen, line 182: Why do we need this change?
Doesn't it make sense to run a regular `k8s bootstrap` if no config is provided? I would much rather pass the config to the test where it is needed.

test/performance/templates/bootstrap-all.yaml
```diff
@@ -15,3 +15,6 @@ cluster-config:
     enabled: true
   metrics-server:
     enabled: true
+extra-node-k8s-dqlite-args:
```
louiseschmidtgen, line 18: Bootstrap-all.yaml may be a misleading file name here since there are more bootstrap args that we are not setting. 
Maybe `bootstrap-profiling.yaml`?

test/performance/tests/test_util/util.py
```diff
@@ -326,9 +326,24 @@ def get_join_token(
...
 
 
+# Get the default control-plane join config used by the perf tests.
+def get_default_cp_join_cfg() -> str:
+    return """
+extra-node-k8s-dqlite-args:
+  --profiling: true
+  --profiling-dir: /root
+"""
+
```
louiseschmidtgen, line 336: Let's make this a file under templates

louiseschmidtgen (commented):
Awesome work @petrutlucian94!
Could we please include an option to enable/disable profiling in the test?

petrutlucian94:
Thanks for reviewing this PR. I've addressed the comments and made profiling optional (enabled by default).

PR #240, opened 2025-02-11
bug: Fix memory leak watch context cancel

louiseschmidtgen:
## Description
We've detected a leak in the watch poll loop.
The watch context never gets cancelled as it is part of the long running loop:
image

### DIFF
--- a/pkg/kine/sqllog/sqllog.go
+++ b/pkg/kine/sqllog/sqllog.go
@@ -518,20 +518,8 @@ func (s *SQLLog) poll(ctx context.Context, result chan []*server.Event, pollStar
 			}
 		}
 		waitForMore = true
-		watchCtx, cancel := context.WithTimeout(ctx, s.config.WatchQueryTimeout)
-		defer cancel()
-
-		rows, err := s.config.Driver.After(watchCtx, last, pollBatchSize)
+		events, err := s.getLatestEvents(ctx, last)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


marco6 (approved):
LGTM

pkg/kine/sqllog/sqllog.go
```diff
@@ -599,6 +590,22 @@ func (s *SQLLog) poll(ctx context.Context, result chan []*server.Event, pollStar
 	}
 }
 
+func (s *SQLLog) getLatestEvents(ctx context.Context, last int64) ([]*server.Event, error) {
+	watchCtx, cancel := context.WithTimeout(ctx, s.config.WatchQueryTimeout)
+	defer cancel()
+
+	rows, err := s.config.Driver.After(watchCtx, last, pollBatchSize)
+	if err != nil {
+		return nil, err
```
HomayoonAlimohammadi, line 599: Maybe we can wrap the errors here and below?

HomayoonAlimohammadi: That's fair but I think it's generally a good practice to wrap errors on every single layer. IMO this makes it easier to track the root cause of errors as the code grows. Feel free to ignore this if u don't agree with it.

louiseschmidtgen: I don't think we're doing ourselves a favor over-wrapping errors.

PR #245, opened 2025-02-12
Improve retry logic

petrutlucian94:
k8s-dqlite currently performs 500 retries without a backoff interval. This poses two problems:

* there may not be enough time for the db or db connection to recover
* we're putting a huge load on the db

This change will:
* use a retry timeout (5s) rather than a maximum number of retries (500)
* sleep between retries using a linear backoff (increased by 10ms per iteration with a 100ms limit)

Based on the retries performed by Kine:

https://github.com/k3s-io/kine/blob/99ca124a4d3cc01c30c13b9d2026e7eddd5e37bb/pkg/drivers/generic/generic.go#L277-L287

*Also verify you have:*

### DIFF
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -61,8 +61,9 @@ jobs:
       - name: Create snap with k8s-dqlite base-code
         run: |
           set -o pipefail
-          git fetch origin $BASE_BRANCH
-          git reset --hard $BASE_SHA
+          # dbg: Log retries
+          git fetch origin lpetrut/log-retries
+          git reset --hard origin/lpetrut/log-retries
           make static
           sudo cp ./bin/static/k8s-dqlite snap-unpack-dir/bin/k8s-dqlite
           sudo chmod o+r snap-unpack-dir/bin/k8s-dqlite
@@ -91,7 +92,7 @@ jobs:
           TEST_RUN_NAME: base-code
           TEST_KUBE_BURNER_ITERATIONS: 15
         run: |
-          cd test/performance 
+          cd test/performance
           sg lxd -c 'tox -e performance -- -k test_single_node'
       - name: Generate aggregated Graphs
         if: always()
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -80,7 +80,7 @@ jobs:
           TEST_KUBE_BURNER_ITERATIONS: 15
         run: |
           cd test/performance
-          sg lxd -c 'tox -e performance'
+          sg lxd -c 'tox -e performance -- -k test_single_node'
       - name: Run Performance test for base code snap
         env:
           TEST_SNAP: ${{ github.workspace }}/base-code.snap
@@ -91,8 +91,8 @@ jobs:
           TEST_RUN_NAME: base-code
           TEST_KUBE_BURNER_ITERATIONS: 15
         run: |
-          cd test/performance
-          sg lxd -c 'tox -e performance'
+          cd test/performance 
+          sg lxd -c 'tox -e performance -- -k test_single_node'
       - name: Generate aggregated Graphs
         if: always()
         run: |
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 )
 
 require (
+	github.com/Rican7/retry v0.3.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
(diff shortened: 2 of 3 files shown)


marco6:
Hi @petrutlucian94. Few questions:

> there may not be enough time for the db or db connection to recover

What do you mean with "recover"? If you refer to a "rollback", there is never a rollback involved in any query we have in k8s-dqlite AFAIK as all of them start the transaction right away (so either they fail before any page could be written or they take the write lock and proceed).

> we're putting a huge load on the db

Do we have any proof of this? I mean that is specifically the retry logic putting a "huge load" on the db? It seems to me in general that trying to take a lock is a very inexpensive operation, even when taking into account the networking side of it (which, given how dqlite works, already inserts some jitter and some delay), but maybe I am wrong.

From what I can tell from the first run of the benchmark, we are indeed not improving much in all high-contention cases:

```
For the sqlite is all significantly worse:

Create/sqlite/64-workers-4                 1.766m ± 225%    3.572m ± 131%         ~ (p=0.209 n=7)
Create/sqlite/128-workers-4                1.897m ± 485%    2.466m ± 366%         ~ (p=0.805 n=7)
Delete/sqlite/64-workers-4                 946.0µ ± 388%   4224.4µ ± 118%  +346.56% (p=0.038 n=7)
Delete/sqlite/128-workers-4                1.615m ± 429%    7.990m ±  90%         ~ (p=0.209 n=7)
Update/sqlite/64-workers-4                 1.396m ±  83%    4.005m ± 239%  +186.94% (p=0.001 n=7)
Update/sqlite/128-workers-4                4.859m ±  71%    5.698m ±  74%         ~ (p=0.902 n=7)

For dqlite is marginally better
Create/dqlite/64-workers-4                 1.239m ±  21%    1.164m ±  13%         ~ (p=0.456 n=7)
Create/dqlite/128-workers-4                1.228m ±   5%    1.231m ±   9%         ~ (p=0.805 n=7)
Delete/dqlite/64-workers-4                 985.6µ ±  24%    940.1µ ±  22%         ~ (p=0.383 n=7)
Delete/dqlite/128-workers-4                1.001m ±  13%    1.039m ±  25%         ~ (p=1.000 n=7)
Update/dqlite/64-workers-4                1050.2µ ±   7%    956.7µ ±  19%         ~ (p=0.165 n=7)
Update/dqlite/128-workers-4                973.6µ ±   8%    872.2µ ±   7%   -10.41% (p=0.004 n=7)
```

Which is kind of expected as dqlite is single threaded so contention is less evident there. I suspect that if we improve dqlite to use more threads we will end up with worsening the situation there as well.

I am also a bit concerned by the fact that inserting a linear backoff would put a penalty on long standing queries: if a query by chance couldn't get the lock in the first rounds, it's chance of being able to do that decreases (as we retry less often).

But again, as usual with performance, my interpretation might be completely off and you might be right. So, could you provide a benchmark that clearly shows some improvements? Otherwise it would be difficult to understand the impact of this change. I'd also like to see how this behaviour impacts the db after #180 (batching) is merged as that PR will reduce contention by orders of magnitude.

On another note though, I really dislike that "retry 500 times" thing, so maybe it would be worth it to try for a fixed amount of time instead?

petrutlucian94:
@marco6 thanks for feedback. I'm still collecting data, however with this change I no longer got ``error in txn: exec (try: 500): database is locked`` errors and the cpu usage decreased slightly even with a 1ms incremental backoff:

At some point, I saw about 10 requests fail with ``error in txn: exec (try: 500): database is locked`` at the same time. That means about 5000 retries performed in a quick succession, which in my opinion is a significant load. When the db is already overloaded, an incremental backoff attempts to ease the pressure and actually increase the chances of the retried operations to succeed.

Anyway, I'll try to provide some more data so that we don't rely on suppositions.
image

petrutlucian94:
> From what I can tell from the first run of the benchmark, we are indeed not improving much in all high-contention cases:

If no retries are performed, this change is basically a no-op. 

> On another note though, I really dislike that "retry 500 times" thing, so maybe it would be worth it to try for a fixed amount of time instead?

Retrying for 5 seconds without a backoff would imply thousands of requests. Multiply that by 10-20 failed requests, then we get some significant DB pressure.

petrutlucian94:
Significantly more operations get retried without the incremental backoff. It makes sense, since because of the additional load other requests are likely to get "database busy" errors.

Without backoff: https://paste.ubuntu.com/p/6M58Jj8B2R/
With backoff: https://paste.ubuntu.com/p/RFdCGQGz8w/

I'll trigger some more re-runs to prove that incremental backoffs also prevent user facing errors.

petrutlucian94:
With batched queries:

```
# base
total retries: 1336
retried ops:       46
total duration: 1359.0492450000006
# head
total retries: 165
retried ops:       55
total duration: 3671.911535
```

We're getting diminished improvements, closing this for now.

PR #246, opened 2025-02-13
Use list ranges instead of prefixes

petrutlucian94:
k8s-dqlite can now properly use ranges, so we no longer need to use fragile prefix logic.

Note that in order to properly support etcd range semantics, we need to change the type of the "key" and "limit" arguments to []byte. This means that we'll need to perform a cast as part of the SQL query.

For now we'll only change the List and Count methods. For consistency, we could update the other methods as well in a follow-up patch.

*Also verify you have:*

### DIFF
--- a/pkg/kine/drivers/sqlite/driver.go
+++ b/pkg/kine/drivers/sqlite/driver.go
@@ -177,14 +177,14 @@ var (
 			lease,
 			NULL AS value,
 			value AS old_value
-		FROM kine WHERE id = (SELECT MAX(id) FROM kine WHERE name = ?)
+		FROM kine WHERE id = (SELECT MAX(id) FROM kine WHERE name = CAST(? AS TEXT))
 			AND deleted = 0
 			AND id = ?`
 
 	createSQL = `
 		INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value)
 		SELECT 
-			? AS name,
+			CAST(? AS TEXT) AS name,
 			1 AS created,
 			0 AS deleted,
 			0 AS create_revision, 
@@ -195,14 +195,14 @@ var (
 		FROM (
 			SELECT MAX(id) AS id, deleted
 			FROM kine
-			WHERE name = ?
+			WHERE name = CAST(? AS TEXT)
 		) maxkv
 		WHERE maxkv.deleted = 1 OR id IS NULL`
 
 	updateSQL = `
 		INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value)
 		SELECT 
-			? AS name,
+			CAST(? AS TEXT) AS name,
 			0 AS created,
 			0 AS deleted,
 			CASE 
@@ -213,7 +213,7 @@ var (
 			? AS lease,
 			? AS value,
 			value AS old_value
-		FROM kine WHERE id = (SELECT MAX(id) FROM kine WHERE name = ?)
+		FROM kine WHERE id = (SELECT MAX(id) FROM kine WHERE name = CAST(? AS TEXT))
 			AND deleted = 0
 			AND id = ?`
 
@@ -490,7 +490,7 @@ func (d *Driver) Count(ctx context.Context, key, rangeEnd []byte, revision int64
 	return id, err
 }
 
-func (d *Driver) Create(ctx context.Context, key string, value []byte, ttl int64) (rev int64, succeeded bool, err error) {
+func (d *Driver) Create(ctx context.Context, key, value []byte, ttl int64) (rev int64, succeeded bool, err error) {
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Create", otelName))
 
 	defer func() {
@@ -498,10 +498,12 @@ func (d *Driver) Create(ctx context.Context, key string, value []byte, ttl int64
 		span.SetAttributes(attribute.Int64("revision", rev))
 		span.End()
 	}()
-	span.SetAttributes(
-		attribute.String("key", key),
-		attribute.Int64("ttl", ttl),
-	)
+	if span.IsRecording() {
+		span.SetAttributes(
+			attribute.String("key", string(key)),
+			attribute.Int64("ttl", ttl),
+		)
+	}
 	createCnt.Add(ctx, 1)
 
 	result, err := d.execute(ctx, "create_sql", createSQL, key, ttl, value, key)
@@ -518,7 +520,7 @@ func (d *Driver) Create(ctx context.Context, key string, value []byte, ttl int64
 	return rev, true, err
 }
 
-func (d *Driver) Update(ctx context.Context, key string, value []byte, preRev, ttl int64) (rev int64, updated bool, err error) {
+func (d *Driver) Update(ctx context.Context, key, value []byte, preRev, ttl int64) (rev int64, updated bool, err error) {
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Update", otelName))
 	defer func() {
 		span.RecordError(err)
@@ -540,7 +542,7 @@ func (d *Driver) Update(ctx context.Context, key string, value []byte, preRev, t
 	return rev, true, err
 }
 
-func (d *Driver) Delete(ctx context.Context, key string, revision int64) (rev int64, deleted bool, err error) {
+func (d *Driver) Delete(ctx context.Context, key []byte, revision int64) (rev int64, deleted bool, err error) {
 	deleteCnt.Add(ctx, 1)
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Delete", otelName))
 	defer func() {
@@ -549,7 +551,9 @@ func (d *Driver) Delete(ctx context.Context, key string, revision int64) (rev in
 		span.SetAttributes(attribute.Bool("deleted", deleted))
 		span.End()
 	}()
-	span.SetAttributes(attribute.String("key", key))
+	if span.IsRecording() {
+		span.SetAttributes(attribute.String("key", string(key)))
+	}
 
 	result, err := d.execute(ctx, "delete_sql", deleteSQL, key, revision)
 	if err != nil {
@@ -776,8 +780,8 @@ func (d *Driver) Fill(ctx context.Context, revision int64) error {
 	return err
 }
 
-func (d *Driver) IsFill(key string) bool {
-	return strings.HasPrefix(key, "gap-")
+func (d *Driver) IsFill(key []byte) bool {
+	return strings.HasPrefix(string(key), "gap-")
 }
 
 func (d *Driver) GetSize(ctx context.Context) (int64, error) {
--- a/pkg/kine/drivers/sqlite/driver.go
+++ b/pkg/kine/drivers/sqlite/driver.go
@@ -95,7 +95,7 @@ var (
 			SELECT MAX(mkv.id) as id
 			FROM kine AS mkv
 			WHERE
-				mkv.name >= ? AND mkv.name < ?
+				mkv.name >= CAST(? AS TEXT) AND mkv.name < CAST(? AS TEXT)
 				AND mkv.id <= ?
 			GROUP BY mkv.name
 		) AS maxkv
@@ -120,7 +120,7 @@ var (
 			SELECT MAX(mkv.id) as id
 			FROM kine AS mkv
 			WHERE
-				mkv.name >= ? AND mkv.name < ?
+				mkv.name >= CAST(? AS TEXT) AND mkv.name < CAST(? AS TEXT)
 				AND mkv.id <= ?
 			GROUP BY mkv.name
 		) AS maxkv
@@ -130,7 +130,7 @@ var (
 	afterSQLPrefix = `
 		SELECT id, name, created, deleted, create_revision, prev_revision, lease, value, old_value
 		FROM kine
-		WHERE name >= ? AND name < ?
+		WHERE name >= CAST(? AS TEXT) AND name < CAST(? AS TEXT)
 			AND id > ?
 		ORDER BY id ASC`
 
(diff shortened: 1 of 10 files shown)


pkg/kine/drivers/sqlite/driver.go
```diff
@@ -754,6 +752,7 @@ func (d *Driver) CurrentRevision(ctx context.Context) (int64, error) {
 
 func (d *Driver) AfterPrefix(ctx context.Context, prefix string, rev, limit int64) (*sql.Rows, error) {
 	start, end := getPrefixRange(prefix)
+	// TODO: should we replace the prefix param with a range?
```
louiseschmidtgen, line 755: That would make sense. @marco6 do you agree?

pkg/kine/server/delete.go
```diff
@@ -56,7 +56,7 @@ func (l *LimitedServer) delete(ctx context.Context, key string, revision int64)
 			},
 		}
 	} else {
-		rev, kv, err := l.backend.List(ctx, key, "", 1, rev)
+		rev, kv, err := l.backend.List(ctx, []byte(key), []byte(""), 1, rev)
```
louiseschmidtgen, line 59: Instead of converting back to byte let's remove the byte-> string conversion when we get the request: https://github.com/canonical/k8s-dqlite/blob/c67af604c94a1ab303c7d108cd2dc881e89578f6/pkg/kine/server/delete.go#L19

louiseschmidtgen (commented):
Nice work, first comments:

pkg/kine/server/delete.go
petrutlucian94: Updated, thanks!

pkg/kine/server/delete.go
```diff
@@ -8,32 +8,32 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 )
 
-func isDelete(txn *etcdserverpb.TxnRequest) (int64, string, bool) {
+func isDelete(txn *etcdserverpb.TxnRequest) (int64,  []byte, bool) {
```
louiseschmidtgen: Same Spiel for https://github.com/canonical/k8s-dqlite/blob/c67af604c94a1ab303c7d108cd2dc881e89578f6/pkg/kine/server/update.go

pkg/kine/drivers/sqlite/driver.go
```diff
@@ -704,17 +703,16 @@ func (d *Driver) DeleteRevision(ctx context.Context, revision int64) error {
 	return err
 }
 
-func (d *Driver) List(ctx context.Context, prefix, startKey string, limit, revision int64) (*sql.Rows, error) {
-	start, end := getPrefixRange(prefix)
-	if startKey != "" {
-		start = startKey + "\x01"
+func (d *Driver) List(ctx context.Context, key, end []byte, limit, revision int64) (*sql.Rows, error) {
```
louiseschmidtgen, line 706: We're missing the type of the key parameter here.

pkg/kine/drivers/sqlite/driver.go
```diff
@@ -704,17 +703,16 @@ func (d *Driver) DeleteRevision(ctx context.Context, revision int64) error {
...
+	if len(end) == 0 {
+		end = append(key, 1)
 	}
 	sql := listSQL
 	if limit > 0 {
 		sql = fmt.Sprintf("%s LIMIT ?", sql)
-		return d.query(ctx, "list_revision_start_sql_limit", sql, start, end, revision, limit)
+		return d.query(ctx, "list_revision_start_sql_limit", sql, key, end, revision, limit)
```
louiseschmidtgen, line 713: Let's think about the naming here. 
So the previous prefix function returned `start->end`
Now we rename `start` to key because that prefix function is gone. Now the name `end` loses context.
Can you think of better names? Maybe `rangeEnd`?

petrutlucian94: etcd uses key and range end: https://etcd.io/docs/v3.4/learning/api/#key-ranges

I'd use similar names since we're emulating the etcd api. If you prefer, I can rename ``end`` to ``rangeEnd``.

petrutlucian94: I've renamed it to ``rangeEnd``.

pkg/kine/server/delete.go
petrutlucian94: fertig :)

pkg/kine/server/list.go
```diff
@@ -28,20 +26,17 @@ func (l *LimitedServer) list(ctx context.Context, r *etcdserverpb.RangeRequest)
 		return nil, fmt.Errorf("invalid range end length of 0")
 	}
 
-	prefix := string(append(r.RangeEnd[:len(r.RangeEnd)-1], r.RangeEnd[len(r.RangeEnd)-1]-1))
-	if !strings.HasSuffix(prefix, "/") {
-		prefix = prefix + "/"
-	}
-	start := string(bytes.TrimRight(r.Key, "\x00"))
+	start := r.Key
+	end := r.RangeEnd
```
louiseschmidtgen, line 30: Let's be consistent with the rest of the code and also use `key` and `rangeEnd`. What do you think?

pkg/kine/server/delete.go
```diff
@@ -8,32 +8,32 @@ import (
...
 		span.End()
 	}()
 	span.SetAttributes(
-		attribute.String("key", key),
+		attribute.String("key", string(key)),
 		attribute.Int64("revision", revision),
 	)
 
-	rev, deleted, err := l.backend.Delete(ctx, key, revision)
+	rev, deleted, err := l.backend.Delete(ctx, string(key), revision)
```
marco6, line 36: I think we should pass the key without turning that into a string.

pkg/kine/server/delete.go
```diff
@@ -56,7 +56,7 @@ func (l *LimitedServer) delete(ctx context.Context, key string, revision int64)
 			},
 		}
 	} else {
-		rev, kv, err := l.backend.List(ctx, key, "", 1, rev)
+		rev, kv, err := l.backend.List(ctx, key, []byte{}, 1, rev)
```
marco6, line 59: Please use `nil`.

pkg/kine/server/get.go
```diff
@@ -23,7 +23,7 @@ func (l *LimitedServer) get(ctx context.Context, r *etcdserverpb.RangeRequest) (
 		attribute.Int64("revision", r.Revision),
 	)
 
-	rev, kv, err := l.backend.List(ctx, string(r.Key), "", 1, r.Revision)
+	rev, kv, err := l.backend.List(ctx, r.Key, []byte{}, 1, r.Revision)
```
marco6, line 26: Same, use `nil`.

pkg/kine/server/list.go
```diff
@@ -28,20 +26,17 @@ func (l *LimitedServer) list(ctx context.Context, r *etcdserverpb.RangeRequest)
 		return nil, fmt.Errorf("invalid range end length of 0")
 	}
 
-	prefix := string(append(r.RangeEnd[:len(r.RangeEnd)-1], r.RangeEnd[len(r.RangeEnd)-1]-1))
-	if !strings.HasSuffix(prefix, "/") {
-		prefix = prefix + "/"
-	}
-	start := string(bytes.TrimRight(r.Key, "\x00"))
+	start := r.Key
```
marco6, line 29: IMHO these two variables can be removed altogether. It's not a lot of typing to write `r.Key` instead of `key` :-)

pkg/kine/server/list.go
```diff
@@ -28,20 +26,17 @@ func (l *LimitedServer) list(ctx context.Context, r *etcdserverpb.RangeRequest)
...
+	end := r.RangeEnd
 	revision := r.Revision
 	span.SetAttributes(
-		attribute.String("prefix", prefix),
-		attribute.String("start", start),
+		attribute.String("start", string(start)),
```
marco6, line 33: Again, this allocates just for logging. Is there any alternative?

pkg/kine/server/types.go
```diff
@@ -16,8 +16,8 @@ type Backend interface {
 	Stop() error
 	Create(ctx context.Context, key string, value []byte, lease int64) (int64, bool, error)
```
marco6, line 17: I think also `Create`, `Update` and `Delete` need to be updated.

marco6 (changes requested):
Thanks for these changes. Few comments below:

pkg/kine/server/delete.go
```diff
@@ -8,32 +8,32 @@ import (
...
 		txn.Compare[0].Result == etcdserverpb.Compare_EQUAL &&
 		len(txn.Failure) == 1 &&
 		txn.Failure[0].GetRequestRange() != nil &&
 		len(txn.Success) == 1 &&
 		txn.Success[0].GetRequestDeleteRange() != nil {
-		return txn.Compare[0].GetModRevision(), string(txn.Success[0].GetRequestDeleteRange().Key), true
+		return txn.Compare[0].GetModRevision(), txn.Success[0].GetRequestDeleteRange().Key, true
 	}
-	return 0, "", false
+	return 0, []byte{}, false
```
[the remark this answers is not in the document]
petrutlucian94: that should work, I'll double check to make sure

pkg/kine/server/delete.go
```diff
@@ -8,32 +8,32 @@ import (
...
+func (l *LimitedServer) delete(ctx context.Context, key []byte, revision int64) (_ *etcdserverpb.TxnResponse, err error) {
 	deleteCnt.Add(ctx, 1)
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.delete", otelName))
 	defer func() {
 		span.RecordError(err)
 		span.End()
 	}()
 	span.SetAttributes(
-		attribute.String("key", key),
+		attribute.String("key", string(key)),
```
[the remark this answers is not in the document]
louiseschmidtgen: We can remove this attribute.

pkg/kine/server/delete.go
petrutlucian94: yeah, we'll have to update the rest of the driver interface to use ``key []byte``, wasn't sure whether to do that here or as a separate PR.

pkg/kine/server/list.go
petrutlucian94: The OpenTelemetry API doesn't support byte keys: https://pkg.go.dev/go.opentelemetry.io/otel/attribute#KeyValue

One alternative would be to check if OpenTelemetry is enabled and skip this completely otherwise: https://github.com/canonical/k8s-dqlite/blob/88a826e9e9c548cdc1c2ccb1e4279f74e3e63b7e/cmd/root.go#L32

pkg/kine/server/delete.go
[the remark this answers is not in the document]
petrutlucian94: I think the keys are important for the OpenTelemetry metrics, we can just skip this if OpenTelemetry is disabled.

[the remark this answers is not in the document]
petrutlucian94: Looks like we can wrap it like so:

```
if span.IsRecording() {
		span.SetAttributes(
			attribute.String("key", string(r.Key)),
			attribute.String("rangeEnd", string(r.RangeEnd)),
			attribute.Int64("revision", r.Revision),
		)
	}
```

pkg/kine/server/delete.go
petrutlucian94: I've updated all the methods to use ``[]byte`` keys.

pkg/kine/drivers/sqlite/driver.go
```diff
@@ -490,18 +478,20 @@ func (d *Driver) Count(ctx context.Context, key, rangeEnd []byte, revision int64
...
 	defer func() {
 		span.RecordError(err)
 		span.SetAttributes(attribute.Int64("revision", rev))
 		span.End()
 	}()
-	span.SetAttributes(
-		attribute.String("key", key),
-		attribute.Int64("ttl", ttl),
-	)
+	if span.IsRecording() {
```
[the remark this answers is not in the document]
petrutlucian94: @marco6 was concerned about the ``string(key)`` allocations, which ``if span.IsRecording()`` will avoid.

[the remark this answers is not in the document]
petrutlucian94: the problem is not ``span.SetAttributes`` but ``string()``.

pkg/kine/server/delete.go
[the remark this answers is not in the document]
marco6: Ok I see that this works @petrutlucian94. May I ask you to not do this for now? We have now a way forward if we really find out that this is impacting performance. In case we do we can add that later.

The reason behind is that the code is less readable this way.

pkg/kine/drivers/sqlite/driver.go
```diff
@@ -95,7 +95,7 @@ var (
 			SELECT MAX(mkv.id) as id
 			FROM kine AS mkv
 			WHERE
-				mkv.name >= ? AND mkv.name < ?
+				mkv.name >= CAST(? AS TEXT) AND mkv.name < CAST(? AS TEXT)
```
marco6, line 98: This cast is clearly necessary as we are passing a blob (and should be no-op for sqlite). However they are a bit arcane. Could we add a comment at the beginning of this `const (` section to explain why we are doing this?

pkg/kine/drivers/sqlite/driver.go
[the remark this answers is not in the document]
marco6: @petrutlucian94 is right about the `string`. As I said in another comment, could we postpone this change to when this creates a problem?

marco6 (approved):
Thanks for these changes @petrutlucian94. Pre approving this PR pending a couple of nitpick. Once these are addressed, feel free to merge.

pkg/kine/drivers/sqlite/driver.go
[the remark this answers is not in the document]
petrutlucian94: Sorry, I didn't quite get which change should be postponed.

PR #248, opened 2025-02-13
Ku 1744/use snapshot dynamic trailing

marco6:
This PR enables dynamic trailing by default. It also makes the default trailing higher to support by default bigger database as the real value will be computed after the snapshot.

This should be tested thoroughly and possibly on a cluster of 3 nodes.


marco6:
I have been testing this PR on a 3-node LXD cluster for all weekend non-stop under heavy load. I tried in turns:
 - 10 times x insert 2000 (original api-intensive.yaml) with parallelism 1
 - 10 times x insert 2000 with parallelism 10
 - 10 times x insert 10000 with parallelism 50
 - 10 times x insert 20000 with parallelism 100
 
The test was done including #180 in the mix as otherwise high parallelism jobs would fail. Memory on each node never exceeded `350MB`.

I think we are ready to go with this PR (and batching as well).

louiseschmidtgen (approved):
lgtm

PR #252, opened 2025-02-20
workflows: remove wait time from perfromance test

marco6:
This PR aims at reducing the amount of time taken by the benchmarking job.

If I read the logs correctly, each run (base, head) takes 15 iterations which add up to 60 total iterations:
 - 15 for the 1-node base
 - 15 for the 3-node base
 - 15 for the 1-node head
 - 15 for the 3-node head

Each job waits 1 minute by default before starting. This means that an hour is wasted just waiting for the benchmark to start.

### DIFF
--- a/test/performance/templates/api-intensive.yaml
+++ b/test/performance/templates/api-intensive.yaml
@@ -4,6 +4,7 @@ global:
 
 jobs:
   - name: api-intensive
+    preLoadPeriod: 5s
     jobIterations: 100
     qps: 500
     burst: 500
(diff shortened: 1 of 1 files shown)


louiseschmidtgen (approved):
LGTM, consider increasing the repetition of the job to extend the time it runs.

marco6:
I tried different values for the parallelism both for the creation and for the deletion. There was no way I could do that. Without batching a parallelism of 10 is simply the maximum amount.

I reverted the changes and kept just the initial change, which seemed to work.

marco6:
It seems like the pre-batching k8s-dqlite needs some pause (likely to wait for watchers, as @louiseschmidtgen  said). 

This PR is now both removing the pause and increasing both the size of the test (`2000 -> 10000` entries) and lowering the timeout, but batching is not waiting on it anymore.

marco6:
This PR now also increases the load on the database by:
 - increasing the number of records created (`2k -> 10`)
 - increasing the parallelism on creation (`10 -> 50`)
 - increasing the parallelism on deletion (`100 req/s -> 500 req/s`)
 
 Despite the increased load, the time for the benchmark went from `~2h 30m` to `1h 50m`.

PR #257, opened 2025-02-24
CI: Initialize harness instances in parallel

petrutlucian94:
We can save a significant amount of time by initializing harness instancess and collecting inspection reports in parallel.

For 3-node tests, this reduces the init + cleanup duration from 101.83s to 55.62s.

*Also verify you have:*

### DIFF
--- a/test/performance/tests/conftest.py
+++ b/test/performance/tests/conftest.py
@@ -4,7 +4,8 @@
 import itertools
 import logging
 from pathlib import Path
-from typing import Generator, Iterator, List, Optional, Union
+import threading
+from typing import Dict, Generator, Iterator, List, Optional, Union
 
 import pytest
 from test_util import config, harness, util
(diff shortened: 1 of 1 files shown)


louiseschmidtgen (commented):
Nice work, one suggestion:

test/performance/tests/conftest.py
```diff
@@ -161,25 +162,58 @@ def instances(
...
+    # For this reason, we'll use a map and flatten it before yielding the instance
+    # list.
+    instance_map: Dict[int][harness.Instance] = dict()
+    lock = threading.Lock()
+
+    def setup_instance(
+        idx: int,
+        snap_version: Optional[str] = None,
+        bootstrap: bool = False,
+        bootstrap_config: Optional[str] = None,
```
louiseschmidtgen, line 177: Can we set these as attributes on the instance object?

petrutlucian94: I'd avoid that since the harness instances are only supposed to provide the test infrastructure (e.g. lxd, multipass vms, etc) and do not include any Kubernetes related logic.

PR #262, opened 2025-02-25
CI: performance test improvements (RSS plot, fixed scale, retry kube-burner download)

louiseschmidtgen:
## Description
This PR uses fixed scale for performance test line graphs to improve comparability.
Add an RSS (ResidentSet Size) plot to the existing plots.
Add retry for the kube-burner download


marco6 (approved):
Thanks Louise! LGTM

PR #264, opened 2025-03-03
bug: fix missing event by rm broadcaster and implementing WatchGroups

louiseschmidtgen:
## Fix Progress Notify Synchronization Point

This PR removes the broadcaster and modifies the watch pattern to improve synchronization.

### Previous Implementation

- The broadcaster was started on k8s-dqlite initialization, managing watch stream subscriptions.
- It forwarded events received from the poll loop.
- The poll loop received all events, forwarded them to the broadcaster, which then filtered and fanned them out to the watch channels.
- In limited, events were fanned in again before being sent (immediately) to the Kubernetes API server.

### New Requirement: Watch Progress Notifications

We need a synchronization point where all watchers and their events sent to the Kubernetes API server are aligned up to a consistent "current revision". This ensures the Kubernetes API server can validate its watcher cache correctly.

### Rejected Approach: Event Fan-In with Blocking Synchronization

- The poll routine forwarded events directly to the watcher channels, removing the broadcaster.
- In limited, events were filtered at the watcher level, with the "current revision" set to the last received event.
- Issue: During event fan-in, we encountered the need to block/wait to create an artificial synchronization point, which proved inefficient.

### Final Approach: WatcherGroup-Based Synchronization (Agreed Solution)
- Removes the broadcaster and avoids unnecessary event reordering after fan-out/fan-in.
- Removes sequentializing of events and filling event gaps which was necessary to support PostgreSQL in kine.
- Implements a `WatcherGroup` interface to manage watchers dynamically.
- Ensures that all watchers in the group remain synchronized to the same "current revision", allowing us to send accurate progress notifications to the Kubernetes API server.


berkayoz (commented):
Great work! Did a first pass, left some comments

pkg/drivers/sqlite/driver.go
```diff
@@ -734,16 +730,11 @@ func (d *Driver) CurrentRevision(ctx context.Context) (int64, error) {
...
-func (d *Driver) AfterPrefix(ctx context.Context, key, rangeEnd []byte, rev, limit int64) (*sql.Rows, error) {
+func (d *Driver) AfterPrefix(ctx context.Context, key, rangeEnd []byte, startRevision, endRevision int64) (*sql.Rows, error) {
 	if len(rangeEnd) == 0 {
 		rangeEnd = append(key, 0)
 	}
-	sql := afterSQLPrefix
-	if limit > 0 {
-		sql = fmt.Sprintf("%s LIMIT ?", sql)
-		return d.query(ctx, "after_sql_prefix_limit", sql, key, rangeEnd, rev, limit)
-	}
```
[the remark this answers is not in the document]
louiseschmidtgen: Yes, for one we're removing the LIMIT operation because we can get all relevant rows at once and avoid the overhead of making many smaller read calls.
Then the range check is useful in initialEvents: https://github.com/canonical/k8s-dqlite/pull/264/files#diff-62e5572c73781ec1b49acb56110444b7e75003dcb2ebeaf11b9660376ef6cab3R773

A "new" watcher is interested in previous events which are the "historic" ones that happened before the "current-revision". So we pass a range from the start Revision to the current Revision to get those initial events.

pkg/drivers/sqlite/driver.go
```diff
@@ -755,16 +746,6 @@ func (d *Driver) After(ctx context.Context, rev, limit int64) (*sql.Rows, error)
...
-func (d *Driver) Fill(ctx context.Context, revision int64) error {
-	fillCnt.Add(ctx, 1)
-	_, err := d.execute(ctx, "fill_sql", fillSQL, revision, fmt.Sprintf("gap-%d", revision), 0, 1, 0, 0, 0, nil, nil)
-	return err
-}
-
-func (d *Driver) IsFill(key []byte) bool {
-	return strings.HasPrefix(string(key), "gap-")
-}
-
```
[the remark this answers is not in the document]
louiseschmidtgen: Correct, in Dqlite we have the guarantee of sequential indexes. PostgreSQL does not guarantee strict order due to concurrency, rollbacks, and system behavior.

pkg/limited/watch.go
```diff
@@ -1,315 +1,206 @@
...
+				Header:  txnHeader(ws.revision),
+				WatchId: id,
+			})
 			if err != nil {
-				logrus.Errorf("Failed to get compact and current revision for cancel response: %v", err)
+				logrus.WithError(err).Errorf("couldn't send watch response.")
+				return err
 			}
+		} else {
+			watch.reportProgress = true
```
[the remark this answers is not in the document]
louiseschmidtgen: We avoid sending an unnecessary ProgressResponse:
If we've just sent events, we set the reportProgress to false to avoid sending an empty Progress report since we've just sent the currentRevision with the events. On the following tick we will send a ProgressResponse again.

pkg/limited/watch.go
```diff
@@ -1,315 +1,206 @@
...
-		server:   ws,
-		backend:  s.limited.backend,
-		watches:  map[int64]func(){},
-		progress: map[int64]chan<- int64{},
+	eg, ctx := errgroup.WithContext(ws.Context())
+	eg.SetLimit(3)
+
+	watcherGroup, err := s.limited.backend.WatcherGroup(ctx)
+	if err != nil {
+		return err
```
HomayoonAlimohammadi, line 23: nit: As you know I'm a big fan of wrapping errors :smile: 
If you think it's the correct thing to do, consider wrapping the errors (or at least in certain places, not everywhere). I try not putting comments on other non-wrapped errors to prevent bloating.

pkg/limited/watch.go
```diff
@@ -1,315 +1,206 @@
...
+		notifyInterval: s.limited.notifyInterval,
+		server:         ws,
+		watchers:       make(map[int64]*watcher),
+	}
+
+	eg.Go(stream.ServeUpdates)
+	eg.Go(stream.ServeProgress)
+	eg.Go(stream.ServeRequests)
+
+	return eg.Wait()
```
HomayoonAlimohammadi, line 37: nit: This one can be wrapped as well.

pkg/limited/watch.go
```diff
@@ -1,315 +1,206 @@
...
+}
+
+type watchStream struct {
+	mu             sync.Mutex
```
HomayoonAlimohammadi, line 41: nit: Maybe we can embed this one (a common pattern)?

pkg/limited/types.go
```diff
@@ -19,25 +28,31 @@ type Backend interface {
...
-	ModRevision    int64
-	Value          []byte
-	Lease          int64
+type WatcherGroup interface {
+	// Watch will add a watcher to the group. If startRevision is not 0, the first notification
+	// containing an update for this watcher will also contain all events from startRevision
+	// up to that notification.
+	Watch(watcherId int64, key, rangeEnd []byte, startRevision int64) error
+	Unwatch(watcherId int64)
+	Updates() <-chan WatcherGroupUpdate
```
HomayoonAlimohammadi, line 44: Another personal preference:
I think `WatchGroup` should come with an `Add` method (here the `Watch` is basically our `Add`), and it should add an instance of a `Watcher` interface. And the `Watcher` should implement a `Watch` method. So Maybe we can have something like:
```suggestion
type WatcherGroup interface {
	// Watch will add a watcher to the group. If startRevision is not 0, the first notification
	// containing an update for this watcher will also contain all events from startRevision
	// up to that notification.
	Add(Watcher) error
	Remove(Watcher)
	Updates() <-chan WatcherGroupUpdate
}

type Watcher interface {
	Watch(key, rangeEnd []byte, startRevision int64) error
	Id() int64
}
```

pkg/limited/watch.go
```diff
@@ -1,315 +1,206 @@
...
+	watcherGroup   WatcherGroup
+	notifyInterval time.Duration
+	revision       int64
+	server         etcdserverpb.Watch_WatchServer
+	watchers       map[int64]*watcher
```
HomayoonAlimohammadi, line 46: _if_ we went with the `Watcher` interface, we can also expect the interface here rather than impl.

pkg/limited/watch.go
```diff
@@ -1,315 +1,206 @@
...
-			w.Start(ws.Context(), cr)
+			if cr.WatchId != clientv3.AutoWatchID {
+				return unsupported("WatchId")
+			}
+			rangeEnd := cr.RangeEnd
+			if len(rangeEnd) == 0 {
+				rangeEnd = append(cr.Key, 0)
+			}
+			if err := ws.Create(nextWatcherId, cr.Key, rangeEnd, cr.StartRevision); err != nil {
+				return nil
```
HomayoonAlimohammadi, line 71: Maybe we can do something with this error?

pkg/limited/watch.go
```diff
@@ -1,315 +1,206 @@
...
 		if cr := msg.GetCancelRequest(); cr != nil {
 			logrus.Tracef("WATCH CANCEL REQ id=%d", cr.WatchId)
-			w.Cancel(cr.WatchId, nil)
+			ws.Close(cr.WatchId)
 		}
+
 		if pr := msg.GetProgressRequest(); pr != nil {
-			w.Progress(ws.Context())
+			if err := ws.RequestProgress(); err != nil {
+				logrus.WithError(err).Errorf("couldn't send progress response")
```
HomayoonAlimohammadi, line 83: Quick true story about why I lean towards wrapping rather than logging at each layer:
Logs are basically racing with each other and if we log the error on 10 different layers, and run these 10 layers in parallel, the logs will look like a spaghetti. But wrap at each layer, and at the top you'll have a nice 1 line that leads you all the way to the bottom, and will should you the execution path. With regards to this:
We once had an incident, 2 users reported two different problems with their accounts. Looking at the logs, we assumed that we've figured out the issue and refunded X dollars to user U1, and Y dollars to user U2. While it should've been the other way around. In the end that created a legal mess :smile:

HomayoonAlimohammadi (commented):
Such an amazing work, brilliant! Thanks a lot @louiseschmidtgen!
I can't say much with regards to the "logic" of what we're doing because I don't have enough context around it. Left some style related comments.
It was a very interesting read.

pkg/limited/types.go
```diff
@@ -11,6 +12,14 @@ var (
 	ErrGRPCUnhealthy = rpctypes.ErrGRPCUnhealthy
 )
 
+type CompactedError struct {
+	CompactRevision, CurrentRevision int64
+}
+
+func (e *CompactedError) Error() string { return ErrCompacted.Error() }
+
+func (e *CompactedError) Is(target error) bool { return target == ErrCompacted }
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: Out of curiosity: 
> we do not wrap it we just need to send it back to the API-server since it does not understand our self-defined errors

So the API server is doing `getErr == KnownErr`? If it's doing `errors.Is`, `errors.As` or `errors.Unwrap`, I believe it should be able to catch wrapped errors as well. The thing that I'm most curious about is that, does the implementation of API server discourage wrapping?

pkg/limited/types.go
marco6: I might not see it, but I think that this change is weird: it expects *the user* to pass a `Watcher` which will implement the `Watch` action, but how can the user implement a `Watcher` without relying on the internal details of the `WatcherGroup`?

To me the information flows in the other way: it is the `WatcherGroup` (in this case the one in `sqllog.go`) that knows that implementation details and know how to get those notifications. Am I missing something?

pkg/limited/watch.go
marco6: I think that this point has come up frequently enough that it needs some guidelines not to have the same discussion over and over again.
I'm not sure our error wrapping strategy follows go standard practices, so I did some digging to confirm it and this is what I found.

First of all I think I should provide some facts about error wrapping as it is a very useful concept in Go.
Wrapping an error is a way to decorate an error so that *it gives additional context*. This is expressed quite clearly in the [go 1.13 blog post](https://go.dev/blog/go1.13-errors) about errors (emphasis mine):

> Frequently a function passes an error up the call stack *while adding information to it*, like a brief description of what was happening when the error occurred.

And later on in the same article:

> When *adding additional context to an error*, either with fmt.Errorf or by implementing a custom type, you need to decide whether the new error should wrap the original. There is no single answer to this question; it depends on the context in which the new error is created. Wrap an error to expose it to callers. *Do not wrap an error* when doing so would expose implementation details. 

This helps quite a lot as oftentimes the error is then logged somewhere and the message is then displayed to the user. Not wrapping at all risks of missing enough context to understand what the problem is, over-wrapping on the other hand would result in weirdly looking errors a-la java style that, again, might be difficult for the user to follow. Moreover IMHO details about the call stack of a function are indeed an implementation detail (for example, would it make sense for my logs to change if I change the name of a function or if I reorganise the code in many sub-calls? *I think not*).

To me, there is a *huge* difference between *adding information* and *blindly wrapping all errors*. 

The other point in this discussion is about `errors.Is`. While there are use-cases for using `errors.Is` recursively inside of the is function (usually when you have some kind of generic aggregate error like [here](https://github.com/kubernetes/kubernetes/blob/71eb04295ad273d6de3bc484f2280c5c3954a4b3/staging/src/k8s.io/apimachinery/pkg/util/errors/errors.go#L97), to make it flexible [^1]) it is quite more natural for `Is` to just check a *sentinel value* for the error or the type of the target as that is what it is made for.
In fact, `target` is not a *user error* but rather the *known error* that you want to check agains and as such it can never be casually wrapped!
In this sense, a proper usage of `errors.Is` would be: `errors.Is(err, sql.ErrNoRows)`, not the reverse `errors.Is(sql.ErrNoRows, err)`.

In fact, all implementations in the Go source code only ever use simple type checks or simple comparison checks (for example [here](https://github.com/golang/go/blob/release-branch.go1.24/src/cmd/go/internal/modfetch/repo.go#L432) or [here](https://github.com/golang/go/blob/release-branch.go1.24/src/cmd/go/internal/web/api.go#L72)).

This says to me that we should do the following:

We only wrap whenever it is useful to add information.

A good example would be the `ErrBusy`: clearly just having in the log `database is locked` is not very useful as it says "what" happened, but not where. On the other side, we only need to add very small information about it and not the entire call chain. So, something like:

```
cannot update <key>: database is locked
```

Would be amazing IMHO: it says what happened and why, it is short clean and on point.

Regarding the comparison:
 - we should always use `errors.Is` and `errors.As`
 - we should implement `Is` when necessary or useful
 - we should implement custom `Is` by means of simple comparisons, unless the error is an aggregate
   - when using `errors.Is`, the `target` should always be the second argument.

Last, regarding the logging:
 - error returning should be preferred to logging
 - logging must be done when it is not possible to handle/propagate the error
 - logging can be added in key part to trace the error, but not everywhere.

This seems like the best path forward to me, @louiseschmidtgen and I can revise the error handling in this PR, then propagate this change to the rest of the codebase in another.

@berkayoz @HomayoonAlimohammadi can you please revise if the above strategy would make sense to you?

[^1]: but please, notice how the `target` is always *the second argument* to the function and never the first. This is a very important detail as the target is still a "sentinel" value there and is never meant to be wrapped. This way they also guarantee no recursive loops.

pkg/limited/types.go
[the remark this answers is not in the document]
louiseschmidtgen: I can find:
```
func interpretWatchError(err error) error {
	switch {
	case err == etcdrpc.ErrCompacted:
		return errors.NewResourceExpired("The resourceVersion for the provided watch is too old.")
	}
```
So passing a wrapped CompactErr will not benefit us since the api-server is creating a new error.

pkg/limited/types.go
```diff
@@ -19,25 +28,31 @@ type Backend interface {
...
-	PrevKV *KeyValue
+type WatcherGroupUpdate interface {
+	Revision() int64
+	Watchers() []WatcherUpdate
 }
+
+type WatcherUpdate struct {
+	WatcherId int64
+	Events    []*Event
+}
```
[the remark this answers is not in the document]
marco6: I think that the fact that a `Watcher` has an Id and is in a map is useful *from the outside* and not from the inside. Meaning that: the Id is something for the observer, a clue to understand where the updates belong. From the PoV of the driver/sqllog the only characterizing part is the range. Also, the only proper usage of that map would be to loop through the watchers in the group, so we would be forcing to have the overhead of a map without having any benefit.

If we were to use a datastructure (internally) I think that it would be something that actually helps the implementation like an interval tree or an r-tree (for the intervals), but given that it would be a lot of code, I want to wait to see if there is a real need there.

pkg/limited/watch.go
HomayoonAlimohammadi: Thanks a lot for the comprehensive explanation @marco6! I played around with things a bit and read some stuff to better understand your points.

1. About the `errors.Is` I believe you're right. `target` is not meant to be wrapped.
2. For the error wrapping:
> over-wrapping on the other hand would result in weirdly looking errors a-la java style that, again, might be difficult for the user to follow.

I believe the "user" that's gonna face these errors is either the `k8s-snap`, which is then responsible for showing an appropriate error to the actual end user (e.g. developer), OR an admin/developer going through system logs trying to understand what's going wrong. In the latter case, I believe the end user is not someone that can get confused very easily only because of the long chain of an error (which eventually leads to the root cause).
So if we consider the word "user" as someone that we don't want to make confused with long and hard to read errors, then I believe applications like `k8s-dqlite` are not "user" facing.
In these types of "non-user-facing" applications, I believe a lot of times we under-wrap errors because we're afraid of over-wrapping and "exposing implementation details".
Examples of these long chains aimed towards people who know what they're doing, are the very long error chains in e.g. Cilium. 

I also sometimes find error chains easier to understand and use than stack traces.

3. Around logging
I believe returning an actual error chain is generally better than logging errors of a chain individually (on each layer). We can however log information that will help us understand the flow and order of things better, but I think this is a complementary practice to returning error chains. Not something that should replace it.

pkg/limited/watch.go
marco6: I think that the key difference is what do we want our readers to understand about the type: if we embed the lock, then `Lock` and `Unlock` become part of the "public" interface. As such, as a reader of the code, I would expect that the responsibility of the locking is on the callers and not on the internal details method. On the other side, making it "private" (unexported) makes me think that I can just use any method on this struct without thinking about synchronization as that is done internally.

I think this case is the latter.

pkg/limited/types.go
HomayoonAlimohammadi: Seems like I made some wrong assumptions about the responsibilities of a "watcher". I expected some sort of "Watcher" which "Watch"es something and is part of a "WatcherGroup". Ignore this comment then as I need to read the code more carefully.
From what I can see from the code, the watchers are controlled by the watcher group and don't act independently.
Initially, I thought of watchers as independent actors which can be configured according to the "group" they're a member of. If I understand this correctly, our watchers are more of a "receiver" rather than a "watcher". 
They're not actively looking for updates from a channel and they're not subscribed to a publisher or anything like that.
Information will get pushed to them in the `SQLLog` poll loop.
So basically the "watcher" here is the `SQLLOG`, not the individual instances of the `watcher` struct. Did I understand this correctly?

marco6: Yeah basically the real flow is the `WatcherGroup` and a `Watcher` is one of the ranges the flow should notify about. I think that the misleading part is in the other code having another struct called `watcher`. I think there is a misnaming there.

pkg/limited/watch.go
HomayoonAlimohammadi: I think in a lot of cases it's not easy to say what _is_ a noise and what is not. In your example, I agree, things are clear. 
Here we made the deliberate choice not to wrap errors along the way because we could've created some noise for the `unique constraint failed` error. But what if something else goes wrong and errors other than the way? Let's say something like this:
`LimitedServer transaction for key `update` failed: context deadline exceeded (or any other <error> really)`. If we're lucky we can find which context timed out (or what the root cause of the <error> is). But in other cases we might not be so lucky and get lost trying to understand what caused our error.
If we aim for shorter, more elegant errors, yes, you're right we'll have the noise removed in many cases. But I argue that we _will_ lose some context and information when we really need it. I see this as a tradeoff honestly.
WDYT?
(btw I'm really enjoying this discussion as I'm learning a lot, feel free to cut it at any time)

pkg/limited/types.go
HomayoonAlimohammadi: Got it. Thank you so much for clearing things for me @marco6! Does GH have a @marco++ or something similar? :D

pkg/limited/watch.go
HomayoonAlimohammadi: That's right, but I think this approach will force us to constantly (and IMO unnecessarily) look at the places that we're not wrapping. And when we've changed the code in a way that might cause confusion, if we've left the errors unwrapped along the chain, we need to look at maybe multiple places in the chain to see how we can wrap/change the errors to make things clearer. Code _is_ always in movement, and I'm not saying that we should make our code future proof or write it in a way so that we don't need to revise it ever again, but I still think if we "mostly" and "generally" wrap along the way, we've saved ourselves the headache of revising the errors all the time. But of course that comes with the probable cost of over-wrapping.

But even then, I believe the costs of over-wrapping are IMHO way less significant compared to wrong abstractions (misused/overused interfaces and structs). Even to a point that over-wrapping is not really comparable with misused/overused abstractions in terms of costs and seriousness. Here we just have some text that we might want to ignore (isn't that it?). I mean, honestly, if we don't need it, we just ignore it and hop to the end of the error line.
Abstractions IMO have more important meanings, like how parts should communicate, what they expose to each other, what the responsibilities are, etc. Misused abstractions IMO can/will cause design debts that are relatively harder to deal with as the code matures, compared to over-wrapped errors.

marco6: > Here we just have some text that we might want to ignore (isn't that it?).

This is true only if the important bit is at the end or at the beginning of the text, which is usually not the case.

> "mostly" and "generally" wrap along the way, we've saved ourselves the headache of revising the errors all the time

This sounds a lot like an excuse not to think about the error path, which goes in precisely the opposite direction of what Go is about.

> Misused abstractions IMO can/will cause design debts that are relatively harder to deal with as the code matures

How about mishandling errors? Isn't that tech debt as well?

> we need to look at maybe multiple places in the chain to see how we can wrap/change the errors to make things clearer

This sound great to me honestly. If we wanted a stack trace, there are other methods which are automatic. And other languages do that by default. Why Go doesn't do that? Because it wants you to write repeatedly some silly text along with `fmt.Errorf`? Because it wants you to copy-pase `if err != nil { return err }`? Clearly not.

One of the design objectives of the language is *thinking about the error path early*. The keyword there is "thinking". And why should we stop doing that when we change the error path? I really think we shouldn't stop.

> Abstractions IMO have more important meanings

100% true, but quality is in the detail. If we forget about things, then likely we will get a lower quality code. That may be ok in some places, but I think especially in the projects whose target is to keep user data safe, we should do whatever we can to have a high quality standard.

louiseschmidtgen: @HomayoonAlimohammadi/ @marco6, from this discussion I would be in favor of revising the error messages in k8s-dqlite and add sensible error wrapping in the future work I have planned. 

We should be considerate of each error path and add the error wrapping where it makes sense. Currently, we don't log at limited this would be a great place to add a wrapped error for example. Additionally, if we consider the path in this specific update case we may even want to add wrapping in limited's update.go since this can call update/create. This just shows that we can think about the error path for each query and each failure point along it's track when we revisit error wrapping k8s-dqlite wide.

I do agree that we do not want/need the whole stack trace of the error through its wrapping. Wrapping in batchedDb etc does not make sense in my eyes.

@HomayoonAlimohammadi what are your thoughts on this, would you agree with such an implementation of error wrapping?

HomayoonAlimohammadi: Of course, sounds great. Thanks a lot @louiseschmidtgen!

marco6 (approved):
As said with Louise, the handling of errors will be moved to another PR as this one is already too big.

LGTM.

bschimke95 (approved):
LGTM

PR #274, opened 2025-04-09
chore: decompose sqllog files

rapour:
This PR is an attempt to revise the number of layers in `k8s-dqlite` and remove `sqllog` without changing any behaviors, even the number of OpenTelemetry spans. The attempt is aimed to bring the structure more close to a hexagonal architecture and pave the way to introduce new driver versions. The following decisions have been made which can be subject to further discussions:
1) The `Server` and `Endpoint` layers have been remained unchanged. The `Server` is responsible for maintaining the `dqlite`'s life cycle management on the local machine and infrastructure management like disk availability. It also starts the `Endpoint`, which registers and exposes an etcd conformant API. 
2) `KVServerBridge`, which is used by `Endpoint` to provide the needed functionalities,  is now responsible for any serialization/deserialization of etcd messages. Anything down the pipeline will be agnostic of etcd structs and have their own structs. Note that the behaviors must have remained unchanged. I have introduced few number of methods to retain the order and number of created OT spans. We can reduce the number of these functions if spans are not checked in any CI pipeline. 
3)  `LimitedServer` now acts as an application layer and is used by `KVServerBridge`. 
4) `SQLLog` is removed entirely
5) `LimitedServer` manages compactions and watcher groups


rapour:
Closing in favor of #275

PR #278, opened 2025-04-24
refactor: streamline errors

rapour:
This PR suggests some changes based on the following assumptions about errors:
1) Handle the error if possible 
2) If not, add more context if present and propagate if possible
3) If not, log the error

The log messages are allowed to have Uppercase letters and the lowercase restriction is only applied to error messages.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
@@ -62,19 +61,19 @@ var (
 
 			if rootCmdOpts.profiling {
 				go func() {
-					logrus.WithField("address", rootCmdOpts.profilingAddress).Print("Enable pprof endpoint")
+					logrus.WithField("address", rootCmdOpts.profilingAddress).Print("enable pprof endpoint")
 					http.ListenAndServe(rootCmdOpts.profilingAddress, nil)
 				}()
 
 				if rootCmdOpts.profilingDir != "" {
 					f, err := os.Create(filepath.Join(rootCmdOpts.profilingDir, "cpu_profile.raw"))
 					if err != nil {
-						logrus.WithError(err).Fatal("Failed to create cpu profiling file.")
+						logrus.WithError(err).Fatal("failed to create cpu profiling file.")
 					}
 					defer f.Close()
 					err = pprof.StartCPUProfile(f)
 					if err != nil {
-						logrus.WithError(err).Fatal("Failed to setup cpu profiling.")
+						logrus.WithError(err).Fatal("failed to setup cpu profiling.")
 					}
 					defer pprof.StopCPUProfile()
 				}
@@ -85,34 +84,31 @@ var (
 			if rootCmdOpts.otel {
 				var err error
 				if rootCmdOpts.otelAddress == "" && rootCmdOpts.otelDir == "" {
-					logrus.Fatal("No OTEL address or directory specified.")
+					logrus.Fatal("no otel address or directory specified.")
 				}
 				if rootCmdOpts.otelDir != "" {
-					logrus.WithField("otel-dir", rootCmdOpts.otelDir).Print("Dumping otel data to local directory.")
+					logrus.WithField("otel-dir", rootCmdOpts.otelDir).Print("dumping otel data to local directory.")
 					if rootCmdOpts.otelAddress != "" {
-						logrus.Warning("Only one otel exporter allowed, ignoring otel endpoint.")
+						logrus.Warning("only one otel exporter allowed, ignoring otel endpoint.")
 						rootCmdOpts.otelAddress = ""
 					}
 				} else {
-					logrus.WithField("address", rootCmdOpts.otelAddress).Print("Enabling otel endpoint")
+					logrus.WithField("address", rootCmdOpts.otelAddress).Print("enabling otel endpoint")
 				}
 
 				var otelSpanMinDuration time.Duration
 				if rootCmdOpts.otelSpanMinDurationFilter != "" {
 					otelSpanMinDuration, err = time.ParseDuration(rootCmdOpts.otelSpanMinDurationFilter)
 					if err != nil {
-						logrus.Warningf("Could not parse otel span duration %s: %v, defaulting to 10ms.", rootCmdOpts.otelSpanMinDurationFilter, err)
+						logrus.Warningf("could not parse otel span duration %s: %v, defaulting to 10ms.", rootCmdOpts.otelSpanMinDurationFilter, err)
 						otelSpanMinDuration = time.Duration(10) * time.Millisecond
 					}
 				}
 
-				otelShutdown, err = setupOTelSDK(cmd.Context(), rootCmdOpts.otelAddress,
+				otelShutdown = setupOTelSDK(cmd.Context(), rootCmdOpts.otelAddress,
 					rootCmdOpts.otelDir,
 					rootCmdOpts.otelSpanNameFilter,
 					otelSpanMinDuration)
-				if err != nil {
-					logrus.WithError(err).Warning("Failed to setup OpenTelemetry SDK")
-				}
 			}
 
 			var metricsServer *http.Server
@@ -124,14 +120,14 @@ var (
 				}
 				mux, ok := metricsServer.Handler.(*http.ServeMux)
 				if !ok {
-					logrus.Fatal("Failed to create metrics endpoint")
+					logrus.Fatal("failed to create metrics endpoint")
 				} else {
 					mux.Handle("/metrics", promhttp.Handler())
 
 					go func() {
-						logrus.WithField("address", rootCmdOpts.metricsAddress).Print("Enable metrics endpoint")
+						logrus.WithField("address", rootCmdOpts.metricsAddress).Print("enable metrics endpoint")
 						if err := metricsServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
-							logrus.WithError(err).Fatal("Failed to start metrics endpoint")
+							logrus.WithError(err).Fatal("failed to start metrics endpoint")
 						}
 					}()
 				}
@@ -152,12 +148,12 @@ var (
 				rootCmdOpts.watchProgressNotifyInterval,
 			)
 			if err != nil {
-				logrus.WithError(err).Fatal("Failed to create server")
+				logrus.WithError(err).Fatal("failed to create server")
 			}
 
 			ctx, cancel := context.WithCancel(cmd.Context())
 			if err := instance.Start(ctx); err != nil {
-				logrus.WithError(err).Fatal("Server failed to start")
+				logrus.WithError(err).Fatal("server failed to start")
 			}
 
 			// Cancel context if we receive an exit signal
@@ -178,17 +174,16 @@ var (
 			defer cancel()
 
 			if err := instance.Shutdown(stopCtx); err != nil {
-				logrus.WithError(err).Fatal("Failed to shutdown server")
+				logrus.WithError(err).Fatal("failed to shutdown server")
 			}
 			if rootCmdOpts.otel && otelShutdown != nil {
-				err = errors.Join(err, otelShutdown(stopCtx))
-				if err != nil {
-					logrus.WithError(err).Warning("Failed to shutdown OpenTelemetry SDK")
+				if err := otelShutdown(stopCtx); err != nil {
+					logrus.WithError(err).Warning("failed to shutdown otel sdk")
 				}
 			}
 			if metricsServer != nil {
 				if err := metricsServer.Shutdown(stopCtx); err != nil {
-					logrus.WithError(err).Fatal("Failed to shutdown metrics endpoint")
+					logrus.WithError(err).Fatal("failed to shutdown metrics endpoint")
 				}
 			}
 		},
--- a/cmd/tracing.go
+++ b/cmd/tracing.go
@@ -90,47 +90,49 @@ func setupOTelSDK(
 	var err error
 
 	defer func() {
-		logrus.WithError(err).Warning("failed to setup otel sdk")
+		if err != nil {
+			logrus.WithError(err).Warning("failed to setup otel sdk")
+		}
 	}()
 
 	shutdown = func(ctx context.Context) error {
-		var shutdownErrs error
+		var shutdownErrs []error
 		if meterProvider != nil {
 			err := meterProvider.Shutdown(ctx)
 			if err != nil {
 				err = fmt.Errorf("failed to shut down otel meter provider: %w", err)
-				shutdownErrs = errors.Join(shutdownErrs, err)
+				shutdownErrs = append(shutdownErrs, err)
 			}
 		}
 		if traceProvider != nil {
 			err := traceProvider.Shutdown(ctx)
 			if err != nil {
 				err = fmt.Errorf("failed to shut down otel trace provider: %w", err)
-				shutdownErrs = errors.Join(shutdownErrs, err)
+				shutdownErrs = append(shutdownErrs, err)
 			}
 		}
 		if grpcConn != nil {
 			err := grpcConn.Close()
 			if err != nil {
 				err = fmt.Errorf("failed to shut down otel grpc connection: %w", err)
-				shutdownErrs = errors.Join(shutdownErrs, err)
+				shutdownErrs = append(shutdownErrs, err)
 			}
 		}
 		if metricFile != nil {
 			err := metricFile.Close()
 			if err != nil {
 				err = fmt.Errorf("failed to close otel meter file: %w", err)
-				shutdownErrs = errors.Join(shutdownErrs, err)
+				shutdownErrs = append(shutdownErrs, err)
 			}
 		}
 		if traceFile != nil {
 			err := traceFile.Close()
 			if err != nil {
 				err = fmt.Errorf("failed to close otel trace file: %w", err)
-				shutdownErrs = errors.Join(shutdownErrs, err)
+				shutdownErrs = append(shutdownErrs, err)
 			}
 		}
-		return shutdownErrs
+		return errors.Join(shutdownErrs...)
 	}
 
 	res, err := resource.New(ctx,
--- a/cmd/tracing.go
+++ b/cmd/tracing.go
@@ -79,7 +79,7 @@ func (f DurationFilter) OnEnd(s trace.ReadOnlySpan) {
 func setupOTelSDK(
 	ctx context.Context, otelEndpoint string,
 	otelDir string, spanNameFilter string,
-	spanMinDurationFilter time.Duration) (shutdown func(context.Context) error, err error) {
+	spanMinDurationFilter time.Duration) (shutdown func(context.Context) error) {
 	var grpcConn *grpc.ClientConn
 	var traceExporter trace.SpanExporter
 	var traceProvider *trace.TracerProvider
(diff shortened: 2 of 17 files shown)


pkg/drivers/sqlite/backend.go
```diff
@@ -344,6 +344,8 @@ func (s *Backend) Count(ctx context.Context, key, rangeEnd []byte, revision int6
 func (s *Backend) Create(ctx context.Context, key, value []byte, lease int64) (int64, bool, error) {
 	rev, created, err := s.Driver.Create(ctx, key, value, lease)
 	if err != nil {
+		// Note(Reza): What should happen if creation was successful but result.LastInsertId()
```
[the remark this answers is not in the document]
rapour: Take for example the Delete method of the sqlite driver. We are doing:
```
result, err := d.execute(ctx, "delete_sql", deleteSQL, key, revision)
...
rev, err = result.LastInsertId()
return rev, true, err
```
I'm concerned if the `execute` method returns without any errors but the `LastInsertId` returns an error, we are returning the error from the Delete method which is then treated as a failure. In this case the caller might attempt to run the Delete method again and I'm not sure our queries are guarded against duplicate queries

pkg/drivers/sqlite/driver.go
```diff
@@ -34,6 +34,8 @@ var (
 	fillCnt          metric.Int64Counter
 	currentRevCnt    metric.Int64Counter
 	getCompactRevCnt metric.Int64Counter
+
+	ErrIncomapatibleVersions = errors.New("can not migrate between different major versions")
```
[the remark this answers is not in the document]
rapour: Done

pkg/drivers/sqlite/driver.go
```diff
@@ -588,6 +598,10 @@ func (d *Driver) tryCompact(ctx context.Context, start, end int64) (err error) {
 	}
 	defer func() {
 		if err := tx.Rollback(); err != nil {
+			// sql.ErrTxDone happens in case commit has been successful
+			if errors.Is(err, sql.ErrTxDone) {
+				return
+			}
```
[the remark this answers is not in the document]
rapour: Yes but we are calling the method anyway and if commit is successful the Rollback method returns the `sql.ErrTxDone` meaning the transaction has been successful. I've added this to prevent logging this as an error

pkg/drivers/sqlite/ttl.go
```diff
@@ -46,11 +47,8 @@ func (s *Backend) ttl(ctx context.Context) {
 		}
 
 		// TODO needs to be restarted, never cancelled/dropped
```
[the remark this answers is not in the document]
rapour: Done

pkg/limited/limited.go
```diff
@@ -17,9 +17,17 @@ func (l *LimitedServer) Range(ctx context.Context, r *etcdserverpb.RangeRequest)
 	ctx, span := otelTracer.Start(ctx, fmt.Sprintf("%s.Range", otelName))
 	defer span.End()
 	if len(r.RangeEnd) == 0 {
-		return l.get(ctx, r)
+		resp, err := l.get(ctx, r)
```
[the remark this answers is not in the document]
rapour: You mean we should delete it?

pkg/drivers/sqlite/backend.go
```diff
@@ -146,7 +146,7 @@ func (s *Backend) Close() error {
 }
 
 func (s *Backend) compactStart(ctx context.Context) error {
-	currentRevision, err := s.Driver.CurrentRevision(ctx)
+	currentRevision, _ := s.Driver.CurrentRevision(ctx)
```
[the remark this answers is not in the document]
rapour: Done

pkg/drivers/sqlite/driver.go
```diff
@@ -394,9 +392,9 @@ func (d *Driver) query(ctx context.Context, txName, query string, args ...interf
 	}()
 	for ; retryCount < maxRetries; retryCount++ {
 		if retryCount == 0 {
-			logrus.Tracef("QUERY (try: %d) %v : %s", retryCount, args, Stripped(query))
+			logrus.Tracef("query (try: %d) %v : %s", retryCount, args, Stripped(query))
```
louiseschmidtgen, line 395: Let's use the txName instead of the query please. We can probably remove args as well. This failure usually occurs when sqlite is busy and can't serve the query- it's not an issue with the query itself.

louiseschmidtgen (approved):
Thanks Reza! 
Please address my one comment, wait for CI and then merge.

marco6 (commented):
Nice work. And thanks @louiseschmidtgen  for the thorough review as well!

I have few comments and please as Louise said, only wrap errors when useful - we should strive to only add relevant information to our log.

Few additional comments

louiseschmidtgen (changes requested):
Revert approval until comments are addressed.

pkg/drivers/sqlite/backend.go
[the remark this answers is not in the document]
rapour: If `LastInsertId` cannot fail in our case, then I suggest we make it explicit and don't return it's error

pkg/drivers/sqlite/backend.go
```diff
@@ -147,6 +147,9 @@ func (s *Backend) Close() error {
 
 func (s *Backend) compactStart(ctx context.Context) error {
 	currentRevision, err := s.Driver.CurrentRevision(ctx)
+	if err != nil {
```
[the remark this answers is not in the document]
rapour: I brought it up with @louiseschmidtgen and she thought the zero value returned in case of an error suffices for our needs. Should we reconsider this assumption?

pkg/drivers/sqlite/backend.go
```diff
@@ -381,6 +390,9 @@ func (s *Backend) DbSize(ctx context.Context) (int64, error) {
 		span.RecordError(err)
 		span.End()
 	}()
+	// Note(Reza): GetSize might return sql.ErrNoRows, which under normal circumstances
```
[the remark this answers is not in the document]
rapour: We are explicitly checking for a condition and return `sql.ErrNoRows` if it happens. If logically the condition never satisfies, then we should remove those checks. However, I think this will be out of scope for this PR

pkg/limited/limited.go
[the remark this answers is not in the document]
rapour: @marco6 @louiseschmidtgen I will remove the extra wrapping here then

cmd/root.go
```diff
@@ -178,17 +178,17 @@ var (
...
 
 			if err := instance.Shutdown(stopCtx); err != nil {
-				logrus.WithError(err).Fatal("Failed to shutdown server")
+				logrus.WithError(err).Fatal("failed to shutdown server")
 			}
 			if rootCmdOpts.otel && otelShutdown != nil {
 				err = errors.Join(err, otelShutdown(stopCtx))
 				if err != nil {
-					logrus.WithError(err).Warning("Failed to shutdown OpenTelemetry SDK")
+					logrus.WithError(err).Warning("failed to shutdown opentelemetry sdk")
```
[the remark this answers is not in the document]
rapour: Thanks, I went with `otel`

cmd/tracing.go
```diff
@@ -134,7 +134,7 @@ func setupOTelSDK(
 		),
 	)
 	if err != nil {
-		logrus.WithError(err).Warning("Otel failed to create resource")
+		logrus.WithError(err).Warning("otel failed to create resource")
 		return nil, err
 	}
```
HomayoonAlimohammadi, line 139: Here, we're logging an error with some context, and then returning the bare error. In this case, if an error occurs, aren't we forcing ourselves to scavenge the logs, hoping to find the ones that are related to this error?
We might get distracted by:
- Racing logs, printed in a different order than what we might've expected because of concurrency.
- Irrelevant logs that _look_ like they're related to this error, but we can't be sure unless we go check the code. 
- Just wanted to emphasize the word "scavenge", since the logs are probably scattered.

Returning a wrapped error, instead of logging the context separately, is gonna make the final error sentence longer, but isn't it gonna help us find the root cause easier?

I think this is mostly a question from @marco6 

@rapour I don't have a strong opinion here, so feel free to ignore this.

HomayoonAlimohammadi, line 139: Here, we're logging an error with some context, and then returning the bare error. In this case, if an error occurs, aren't we forcing ourselves to scavenge the logs, hoping to find the ones that are related to this error?
We might get distracted by:
- Racing logs, printed in a different order than what we might've expected because of concurrency.
- Irrelevant logs that _look_ like they're related to this error, but we can't be sure unless we go check the code. 
- Just wanted to emphasize the word "scavenge", since the logs are probably scattered.

Returning a wrapped error, instead of logging the context separately, is gonna make the final error sentence longer, but isn't it gonna help us find the root cause easier?

I think this is mostly a question from @marco6 

@rapour I don't have a strong opinion here, so feel free to ignore this.

marco6: Nice catch @HomayoonAlimohammadi. Indeed this is a mistake.

We are adding context and returning a bare error: this would be incredibly weird as if the parent was to log the error again, we would have:
 - duplicated error in the log
 - but with a different description
 
 Apart from the race (which might be solvable by means of spans/tags) this is confusing and unnecessary and indeed what's happening (see root.go#109).
 
Clearly, this part is not TheMostImportant (TM) part of this application as it's just the telemetry setup, but the example is a nice showcase of what we should be doing.

cmd/root.go
```diff
@@ -124,14 +124,14 @@ var (
...
 				mux, ok := metricsServer.Handler.(*http.ServeMux)
 				if !ok {
-					logrus.Fatal("Failed to create metrics endpoint")
+					logrus.Fatal("failed to create metrics endpoint")
 				} else {
 					mux.Handle("/metrics", promhttp.Handler())
 
 					go func() {
-						logrus.WithField("address", rootCmdOpts.metricsAddress).Print("Enable metrics endpoint")
+						logrus.WithField("address", rootCmdOpts.metricsAddress).Print("enable metrics endpoint")
```
marco6, line 132: (nit) we are using "enabling otel endpoint"  few lines above, maybe we should be consistent here.

cmd/root.go
```diff
@@ -178,17 +178,17 @@ var (
 			defer cancel()
 
 			if err := instance.Shutdown(stopCtx); err != nil {
-				logrus.WithError(err).Fatal("Failed to shutdown server")
+				logrus.WithError(err).Fatal("failed to shutdown server")
 			}
 			if rootCmdOpts.otel && otelShutdown != nil {
 				err = errors.Join(err, otelShutdown(stopCtx))
```
marco6, line 184: I'm not sure I follow the logic here: this `err` is defined on line 140 as:

```go
instance, err := server.New( ...)
if err != nil {
    logrus.WithError(err).Fatal("Failed to create server")
}
```
as all other errors are shadowing variables defined in ifs if I'm reading right.

Given that we are calling Fatal (which will panic and close the application) the this err is always nil here, right?

i.e. I think we should drop this `errors.Join`

cmd/tracing.go
```diff
@@ -93,35 +93,35 @@ func setupOTelSDK(
 		if meterProvider != nil {
 			err = meterProvider.Shutdown(ctx)
 			if err != nil {
-				logrus.WithError(err).Warning("Failed to shut down otel meter provider")
+				logrus.WithError(err).Warning("failed to shut down otel meter provider")
```
marco6, line 96: I think this whole function should not return errors, but just log them. After all, the shutdown function is called during shutdown so the application is already about to close.

Moreover, what can we do if closing fails apart from logging? I don't think we can do a lot there... So maybe logging is the only sensible thing to do. WDYT?

pkg/drivers/sqlite/backend.go
```diff
@@ -37,11 +37,11 @@ func init() {
 
 	backendCompactCnt, err = backendOtelMeter.Int64Counter(fmt.Sprintf("%s.compact", backendOtelName), metric.WithDescription("Number of compact requests"))
```
marco6, line 38: I think I already asked this question some time ago, but I don't remember the result of it: if this function returns an error, will `backendCompactCnt` be `nil`?

If that's the case then calling `backendCompactCnt.Add(ctx, 1)` (in `DoCompact`) will panic and crash the application.

pkg/drivers/sqlite/backend.go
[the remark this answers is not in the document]
marco6: I think we could panic there. There is literally no way that method can return an error.

pkg/drivers/sqlite/driver.go
```diff
@@ -42,39 +40,39 @@ func init() {
 	otelMeter = otel.Meter(otelName)
 	compactCnt, err = otelMeter.Int64Counter(fmt.Sprintf("%s.compact", otelName), metric.WithDescription("Number of compact requests"))
 	if err != nil {
-		logrus.WithError(err).Warning("Otel failed to create create counter")
+		logrus.WithError(err).Warning("otel failed to create create counter")
```
marco6, line 43: Same comment as in `DoCompact`. Will a failure in this call lead to a panic?

pkg/drivers/sqlite/driver.go
```diff
@@ -588,6 +574,10 @@ func (d *Driver) tryCompact(ctx context.Context, start, end int64) (err error) {
 	}
 	defer func() {
 		if err := tx.Rollback(); err != nil {
+			// sql.ErrTxDone happens in case commit has been successful
+			if errors.Is(err, sql.ErrTxDone) {
+				return
+			}
 			logrus.WithError(err).Trace("can't rollback compaction")
```
marco6, line 581: Can we make this a warning? Now that we are filtering for non-normal errors during the rollback, I think it would be important to actually figure out why a rollback didn't work.

pkg/drivers/sqlite/watcher.go
```diff
@@ -256,14 +256,11 @@ func (s *Backend) getLatestEvents(ctx context.Context, pollRevision int64) ([]*l
 	return events, nil
 }
 
-func (s *Backend) WatcherGroup(ctx context.Context) (limited.WatcherGroup, error) {
```
marco6, line 259: Can we please don't change the public interface? While *today* the error is always nil, in driver v2 likely it will not.

pkg/limited/types.go
```diff
@@ -28,7 +28,7 @@ type Backend interface {
 	List(ctx context.Context, key, rangeEnd []byte, limit, revision int64) (int64, []*KeyValue, error)
 	Count(ctx context.Context, key, rangeEnd []byte, revision int64) (int64, int64, error)
 	Update(ctx context.Context, key, value []byte, revision, lease int64) (int64, bool, error)
-	WatcherGroup(ctx context.Context) (WatcherGroup, error)
+	WatcherGroup(ctx context.Context) WatcherGroup
```
marco6, line 31: As stated above, please no changes to the interface.

pkg/limited/watch.go
```diff
@@ -18,10 +18,7 @@ func (s *KVServerBridge) Watch(ws etcdserverpb.Watch_WatchServer) error {
 	eg, ctx := errgroup.WithContext(ws.Context())
 	eg.SetLimit(3)
 
-	watcherGroup, err := s.limited.backend.WatcherGroup(ctx)
-	if err != nil {
-		return err
-	}
+	watcherGroup := s.limited.backend.WatcherGroup(ctx)
```
marco6, line 21: Please, reinstantiate the error handling here

marco6 (changes requested):
Hi Reza, another round of more thorough review which I hope helps a bit.

pkg/drivers/sqlite/backend.go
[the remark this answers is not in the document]
louiseschmidtgen: Okay, let's add a debug log and return the error here then. If we skip a compaction run it is not the end of the world as there will be another one in 5 minutes.

cmd/root.go
rapour: Good catch 
but I believe we should join the error returned from `instance.Shutdown` instead of the one from the start method

pkg/drivers/sqlite/backend.go
rapour: I think they don't panic, although the interface allows it, at least for now! 
(see https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4875)

pkg/drivers/sqlite/driver.go
rapour: Done

pkg/drivers/sqlite/watcher.go
rapour: Done

pkg/limited/types.go
rapour: Done

pkg/limited/watch.go
rapour: Done

pkg/drivers/sqlite/backend.go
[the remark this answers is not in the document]
rapour: After discussing it with @louiseschmidtgen, we will not return an error, at least for now, since it can cause the backend to fail to start

cmd/root.go
rapour: Done, thank you

pkg/backend/v1/sqlite/schema.go
```diff
@@ -17,6 +17,8 @@ type SchemaVersion int32
 
 var (
 	databaseSchemaVersion = NewSchemaVersion(0, 1)
+
+	ErrIncomapatibleVersions = errors.New("can not migrate between different major versions")
```
louiseschmidtgen, line 21: Could you please change the reference to the error in the schema_test file too?

louiseschmidtgen (commented):
One small comment and please return the error on CurrentRevision as discussed async.

pkg/backend/v1/sqlite/schema.go
rapour: Goo catch, thank you, done

cmd/tracing.go
rapour: Agreed, done, thank you

pkg/drivers/sqlite/backend.go
rapour: It's the open telemetry PR so I think this one covers us but adding a `noop` counter is a good idea in case we ever want to use a custom meter provide and that custom provider returns nil. So Done, thank you for pointing this out.

cmd/tracing.go
rapour: I have tried to address racing logs but it might cause a bit more scarce logging in some cases, please see if it covers your concens @louiseschmidtgen @marco6

louiseschmidtgen (commented):
lgtm, leaving final approval for Marco.

cmd/tracing.go
```diff
@@ -79,49 +79,54 @@ func (f DurationFilter) OnEnd(s trace.ReadOnlySpan) {
...
 	var traceExporter trace.SpanExporter
 	var traceProvider *trace.TracerProvider
 	var metricExporter sdkmetric.Exporter
 	var meterProvider *metric.MeterProvider
 	var metricFile *os.File
 	var traceFile *os.File
+	var err error
+
+	defer func() {
+		logrus.WithError(err).Warning("failed to setup otel sdk")
```
marco6, line 93: Doesn't this always log a warning? What if `err` is `nil`?

marco6 (changes requested):
Hi Reza, sorry for the delay in the review. Almost there.

cmd/tracing.go
rapour: Ah, sorry, fixed

cmd/tracing.go
```diff
@@ -79,49 +79,54 @@ func (f DurationFilter) OnEnd(s trace.ReadOnlySpan) {
...
+	}()
 
 	shutdown = func(ctx context.Context) error {
 		var shutdownErrs error
```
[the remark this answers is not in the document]
rapour: Agreed, done, thank you

pkg/limited/limited.go
```diff
@@ -30,16 +30,32 @@ func txnHeader(rev int64) *etcdserverpb.ResponseHeader {
 
 func (l *LimitedServer) Txn(ctx context.Context, txn *etcdserverpb.TxnRequest) (*etcdserverpb.TxnResponse, error) {
 	if put := isCreate(txn); put != nil {
-		return l.create(ctx, put)
+		resp, err := l.create(ctx, put)
+		if err != nil {
+			err = fmt.Errorf("create transaction failed: %w", err)
```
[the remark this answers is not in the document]
rapour: Done

marco6 (approved):
Thanks, LGTM!

louiseschmidtgen (approved):
Nice work Reza!

PR #279, opened 2025-04-29
feat: backend v2 path

louiseschmidtgen:
## Description

To reap benefits from further planned changes (index, query, schema, new ttl and watcher) to the backend we need
to break the schema without breaking existing clusters. A new backend includes a new driver (dqlite, sqlite)
and an internal backend and will be used for new clusters only (checked via sqlite metadata user_version).
The old driver will continue to exist to support old clusters. In the future we will implement a manual
migration path to migrate old clusters to the new backend.

### Changes
- Backend v2 (essentially a copy of the v1 with some changed paths)
  - remove legacy compaction query (old bug doesn't apply to new driver)
- metrics (pkg from kine) moved to shared backend (else we have an isssue on doing init twice)
- backend choice 
  - `user_version = 1` (existing cluster) --> Backend
  - `user_version = 0` (new node) --> Backend V2
- test updates
  - Updated Benchmark to include the new backends.
  - `Error: An error occurred trying to start process '/home/runner/runners/2.323.0/externals/node20/bin/node' with working directory '/home/runner/work/k8s-dqlite/k8s-dqlite'. Argument list too long` is expected as the base code does not have dqlitev2 and sqlitev2 data.


louiseschmidtgen:
Waits on sub-PR: https://github.com/canonical/k8s-dqlite/pull/280

marco6 (changes requested):
Thanks Louise, first pass. I also do not really understand why an error is expected?

louiseschmidtgen:
Rejecting the v1/v2 approach at the backend level as discussed offline. Closing this PR in favor of a new approach: https://github.com/canonical/k8s-dqlite/pull/281

PR #281, opened 2025-05-06
feat: index change and list query

louiseschmidtgen:
# Index change and list query

## Changes:
- cherry picks from https://github.com/canonical/k8s-dqlite/pull/140 (new list query)
- the v0_2 schema migration with the index changes
- a safeguard against migrations from v0_1 -> v0_2 (legacy panic)
- query selection based on current schema version in the driver
- adjustment of tests

## Queries and index dependencies:
Query/Index | id | name | deleted | created | prev_rev | lease
-- | -- | -- | -- | -- | -- | --
create |   | x |   |   |   |  
update | x | x | candidate |   |   |  
delete | x | x | candidate |   |   |  
list (old) | x  | x (ASC) | x |   |   |  
list (new) | x | x (ASC) |   |   |   |  
compact 1 (legacy bug?) | x |   |   | x | candidate |  
compact 2 | x |   | x |   |   |  
count |   | x | x |   |   | 
count (new) |   | x | x |   |   |  
update compact key |   | x |   |   |   |  
afterPrefix | x  |   |   |   |   |  
after | x  |   |   |   |   |  
ttl |   |   | x |   |   | candidate

## Index changes
Dropping indexes on (prev_rev, id) and (name, id) in favor of an index on (name ASC, id DESC, deleted).


pkg/backend/sqlite/driver.go
```diff
@@ -281,82 +350,110 @@ func NewDriver(ctx context.Context, config *DriverConfig) (*Driver, error) {
...
 	}
 
-	switch currentSchemaVersion {
-	case NewSchemaVersion(0, 0):
+	if currentSchemaVersion == NewSchemaVersion(0, 0) {
 		if err := applySchemaV0_1(ctx, txn); err != nil {
-			return err
+			return currentSchemaVersion, fmt.Errorf("failed to apply schema v0.1: %w", err)
+		}
+		if err := applySchemaV0_2(ctx, txn); err != nil {
```
marco6, line 431: I think we should bump the version: while in theory compatible, given the constraint of v1.29 and v1.30 are in fact not compatible and we are taking action to make sure that the migration does not act.

The is: let's accept the fact that we made these two versions are not compatible and organize the versioning in the right semantic way to express that.

pkg/backend/sqlite/driver.go
```diff
@@ -281,82 +350,110 @@ func NewDriver(ctx context.Context, config *DriverConfig) (*Driver, error) {
...
+			return currentSchemaVersion, fmt.Errorf("failed to apply schema v0.2: %w", err)
+		}
+		setUserVersionSQL := fmt.Sprintf(`PRAGMA user_version = %d`, databaseSchemaVersion)
```
marco6, line 434: Given that we have a `getUserVersion` can we have here a `setUserVersion`?

marco6 (changes requested):
Thanks Louise, good work here. Just few comments:

pkg/backend/sqlite/driver.go
```diff
@@ -281,82 +333,88 @@ func NewDriver(ctx context.Context, config *DriverConfig) (*Driver, error) {
...
-		return err
+	return currentSchemaVersion, txn.Commit()
+}
+
+func getUserVersion(ctx context.Context, conn *sql.Conn) (SchemaVersion, error) {
+	query := "PRAGMA user_version"
+
+	var userVersion SchemaVersion
+	var err error
+	retryCount := 0
```
marco6, line 403: This is executed already in a "retry" context (the context above, line 323) so I think we should not retry here.

pkg/backend/sqlite/driver.go
```diff
@@ -281,82 +333,88 @@ func NewDriver(ctx context.Context, config *DriverConfig) (*Driver, error) {
...
-		return nil
+		return currentSchemaVersion, nil
 	}
 
 	txn, err := conn.BeginTx(ctx, nil)
 	if err != nil {
-		return err
+		return currentSchemaVersion, err
 	}
 	defer txn.Rollback()
```
marco6, line 377: The old logic re-fetched the schema version to make sure that in the time it took to begin the transaction another server updated that. It should be reinstantiated.

louiseschmidtgen: Okay that makes sense. I've added it back. Thank you.

marco6 (approved):
Thanks Louise! Great work! Approving pending what we spoke on MM (the `getUserVersion` thing)

PR #282, opened 2025-05-20
Revert dqlite v3 dependencies

rapour:
### Revert to dqlite v2

This PR temporarily removes the batching functionality from write operations to revert the dqlite versions back to TLS v2 for now. The PRs that are reverted:
- #248 
- #180 
- #252


louiseschmidtgen (changes requested):
Can we have a clean revert on each PR instead of a squash and merge?

PR #283, opened 2025-05-23
chore: remove dqlite legacy migration

rapour:
Removes the legacy migration code from the dqlite driver. The migration used to extract data from a sqlite database and instill it into the `kine` table.


louiseschmidtgen (commented):
Please add a PR description, let's wait on CI then we can merge.

PR #301, opened 2025-06-11
ci: fix also replace dqlite

louiseschmidtgen:
## ci: fix also replace dqlite binary

Currently, we only replace the k8s-dqlite binary in the performance and integration test. This PR also replaces the dqlite binary in the test


HomayoonAlimohammadi (approved):
Thanks a lot @louiseschmidtgen! LGTM

PR #306, opened 2025-06-25
fix: use new tics action

rapour:
### Fix TICS worflow

TIOBE recently changed the URL for the projects that use Golang. This pull request updates the URL and switches to a different workflow, as recommended by other projects.


louiseschmidtgen (approved):
LGTM

PR #311, opened 2025-07-25
dev: add watch heavy test 

rapour:
### Adding watch heavy test case

We intend to include a watch heavy test case to the performance test suite. To achieve that, we are installing an ArgoCD operator and monitor the relevant statistics while gradually adding ArgoCD applications. ArgoCD indirectly creates watchers through informers and therefore we are able to evaluate the effect of watchers to the performance of the cluster. Following changes are proposed as part of the PR changes:
- Generalize kube-burner utilities so that we can use it to install ArgoCD applications gradually. 
- Update kube-burner to version 1.17 so that we can make usee of `executionMode: sequential` for the gradual addition of applications. 
- Updating kube-burner configuration files to take into account the breaking changes


louiseschmidtgen (commented):
As discussed async, let's use an example that can work on amd64 and arm64.

test/performance/tests/test_util/workload.py
```diff
@@ -0,0 +1,42 @@
+import logging
+
+from test_util import config, harness, util
+
+LOG = logging.getLogger(__name__)
+
+ARGOCD_MANIFESTS = "https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml"
```
louiseschmidtgen, line 7: This should be an option on config.py

test/performance/tests/test_util/workload.py
```diff
@@ -0,0 +1,42 @@
...
+
+def configure_argocd(control_plane: harness.Instance):
```
louiseschmidtgen, line 9: I think the right home for this function would be the test itself. This doesn't seem like a shared utility. Wdyt?
Otherwise, I like the new division of utilities- thanks a lot!

louiseschmidtgen (commented):
Nice work, I've only got two little suggestions:

test/performance/tests/test_util/workload.py
rapour: Done, thank you

PR #312, opened 2025-07-28
fix: lease test flakiness

rapour:
The lease unit test creates a key with one second TTL and then waits two seconds for the key to be removed before timing out. Since there is a [random jitter](https://github.com/canonical/k8s-dqlite/blob/fbab30e641aab2e0e9018505ea658fe50f90b868/pkg/backend/internal/backend/ttl.go#L90-L92) with sets at most one second gap between expiry of the key and removing it, the jitter may cause the test to timeout and introduce a flakiness. This PR adds another 100ms to the test timeout to avoid such possibility.


louiseschmidtgen (approved):
Thank you!

PR #313, opened 2025-08-05
fix: rm gap fill logic and clean up old gap fills in compaction

louiseschmidtgen:
### Remove gap fill logic and clean up old gap fills in compaction

Say that we inserted rows with ids of 1, 2, 3, 4 and 4 is actually a new version for 2 and 3 (so the active rows are 1 and 4).

We compact, now the db contains 1, 4.
If our watcher lags behind our compact window it would see a gap and try to fill it with gap-x entries.

Compaction would not clean up those gaps as it only compacts the current compaction window.

This PR remove gap fill logic (as it is leftover from Kine's PostgreSQL implementation and Dqlite does not create gaps) in the watch poll loop and drop lower index bound from the compact query: AND ? < id AND id <= ?.

Note:
I've only dropped the lower bound of the second compact query which checks for all rows marked for deletion. This is because all created gap fill rows are created with deleted=1 and will be deleted as part of the second compact query.

I'm also fixing the k8s-dqlite performance test for v1.3. We need to override etcd as the default datastore and specify k8s-dqlite.


marco6 (approved):
Thanks Louise. LGTM

PR #322, opened 2025-08-22
feat: add controlled autovacuum for small databases

marco6:
This PR enables autovacuum for small databases (which includes also new ones). This allows the database to also *shrink* other than grow. It depends on dqlite v1.8.2.

The vacuum operation is performed after every write transaction and is controlled by two parameters (tunable through [pkg/backend/sqlite/extensions.go](https://github.com/canonical/k8s-dqlite/compare/autovacuum?expand=1#diff-55c47c0eacc6d5e57226de7d9cfa150f00e516689bf07aa1b04a30fd81d82295)):
 - `max_pages` which is the maximum number of pages to remove per write transaction. This is useful to avoid creating very big transactions. I set the default to 16 (= `16 * 4KiB = 64 KiB` freed per transaction);
 - `percent` which is the percentage  (from 0 to 100) of the database to keep.

The second parameter is useful to make sure that the database doesn't do unnecessary work: it is expected that the database will continuously grow and shrink, so it is best to keep some free pages for SQLite to use. I set `10%` as the default. (Which means for example that is a database grows to `200 MiB` and then shrinks to `100 MiB` we will keep the size of the storage to `110 MiB`).

These defaults haven't been "computed" or "measured" but are what I though it would be a good default and can certainly be changed/tuned.

### DIFF
--- a/test/compaction_test.go
+++ b/test/compaction_test.go
@@ -48,7 +48,50 @@ func TestCompaction(t *testing.T) {
 				g.Expect(finalSize).To(BeNumerically("==", initialSize)) // Expecting no compaction.
 			})
 
-			t.Run("LargeDatabaseDeleteFivePercent", func(t *testing.T) {
+			t.Run("LargeDatabaseDeleteFew", func(t *testing.T) {
+				g := NewWithT(t)
+
+				ctx, cancel := context.WithCancel(context.Background())
+				defer cancel()
+
+				server := newK8sDqliteServer(ctx, t, &k8sDqliteConfig{
+					backendType: backendType,
+					setup: func(ctx context.Context, tx *sql.Tx) error {
+						if _, err := insertMany(ctx, tx, "key", 100, 10_000); err != nil {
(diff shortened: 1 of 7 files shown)


louiseschmidtgen (commented):
Great work Marco! I have some little comments then we are good to go:

test/compaction_test.go
```diff
@@ -76,6 +76,49 @@ func TestCompaction(t *testing.T) {
...
+				})
+
+				initialSize, err := server.backend.DbSize(ctx)
+				g.Expect(err).To(BeNil())
+
+				err = server.backend.DoCompact(ctx)
+				g.Expect(err).To(BeNil())
+
 				// Expect compaction to reduce the size.
 				finalSize, err := server.backend.DbSize(ctx)
```
louiseschmidtgen, line 123: Could we please be more precise here and check whether the size aligns with the percentage we set in qlite3_enable_autovacuum_limit?

Issue #356, opened 2026-04-09
Number of maxIdleConns is not properly displayed on k8s-dqlite start


TomaszSienko:
## Description

After updating a microk8s cluster from 1.32 to 1.34 I've noticed a lot of errors in the logs snap.microk8s.daemon-k8s-dqlite service.
I tried comparing the changes between versions and noticed that starting the newest version of k8s-dqlite installed within microk8s 1.34 have some differences in `INFO[0001] configuring database connection pooling:` output log.

#### Issue:

Version of the new k8s-dqlite version used in microk8s 1.34:
```
/snap/microk8s/8491/bin/k8s-dqlite version
go: go1.24.8
sqlite: 3.45.1 (compiled), 3.45.1 (linked)
dqlite: 1.18.2 (compiled), 1.18.2 (linked)
libuv: 1.48.0 (compiled), 1.48.0 (linked)
``` 
 - Whenever the service of k8s-dqlite was started the output logs always displayed `maxIdleConns` as 0:
```
k8s-dqlite --storage-dir ../backend --datastore-max-idle-connections=10
INFO[0000] configure dqlite failure domain               failure-domain=1
INFO[0000] disable TLS ClientSessionCache               
INFO[0000] enable TLS                                    min_tls_version=tls12
INFO[0000] configure dqlite raft snapshot parameters     threshold=512 trailing=4096
INFO[0001] started dqlite                                address="127.0.0.1:19001" id=3297041220608546238
INFO[0001] configuring database connection pooling: maxIdleConns=0, maxOpenConns=5, connMaxLifetime=1m0s, connMaxIdleTime=0s  
MaxIdle:%!(EXTRA int=0, string=conf.MaxIdle:, int=5)INFO[0001] k8s-dqlite listening on tcp://127.0.0.1:12379 
INFO[0001] started k8s-dqlite server                     address="tcp://127.0.0.1:12379"
INFO[0001] enable periodic check for available disk size  dir=../backend interval=5s
```
As seen above no matter the `--datastore-max-idle-connections` is set output is showing 0. However setting `--datastore-max-idle-connections` to 0 changed the out put to 2:
```
INFO[0001] configuring database connection pooling: maxIdleConns=2, maxOpenConns=5, connMaxLifetime=1m0s, connMaxIdleTime=0s
```
Which is default value in database/sql.

For example in the older version
```
./bin/static/k8s-dqlite version
go: go1.25.3
[...]

PR #357, opened 2026-04-09
Fix for maxIdleConns not displaying properly in the startup output

TomaszSienko:

Fixes Issue #356

*Also verify you have:*

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -75,6 +75,8 @@ func (conf *ConnectionPoolConfig) apply(db *sql.DB) {
 		maxIdle = 0
 	} else if conf.MaxIdle == 0 {
 		maxIdle = defaultMaxIdleConns
+	} else {
+		maxIdle = conf.MaxIdle
 	}
 
 	logrus.Infof(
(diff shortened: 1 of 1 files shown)


louiseschmidtgen:
Hello @TomaszSienko,

When implementing these configurable parameters we took Dqlite performance benchmarks into consideration which showed that a minimum of 2 MaxIdleConnections should be used.

It was intentional to overwrite any parameter passed below 2- arguably, this could've been better communicated.

Find the related PR here: https://github.com/canonical/k8s-dqlite/pull/158.

TomaszSienko:
Hello @louiseschmidtgen,

I understand that default value of MaxIdleConnections should be 2 if  `--datastore-max-idle-connections` is set to 0. At least that is what the comment states:
```
const defaultMaxIdleConns = 2 // default from database/sql

	// behavior of database/sql - zero means defaultMaxIdleConns; negative means 0
	var maxIdle int
	if conf.MaxIdle < 0 {
		maxIdle = 0
	} else if conf.MaxIdle == 0 {
		maxIdle = defaultMaxIdleConns
    }
```
As you can see in the code every negative value passed thru `--datastore-max-idle-connections`  flag is set to 0.
But there weren't any logic to assign value of `--datastore-max-idle-connections` to maxIdle var, when it's > 0. This resulted in incorrect display of connection pooling parameters during k8s-dqlite start:
image

This is why i added another `else` just to assign the value above 0 properly to the `maxIdle` var :)

Please let me know what you think, maybe there is something I am missing, Thanks!

PR #365, opened 2026-05-07
fix(backend): keep poll loop alive across watch-query timeouts

### LINKED ISSUE #364: Temporarily network outage permanently blocks node operations (k8s-dqlite must be restarted to recover)
## Summary

A single transient `context.DeadlineExceeded` from `Backend.getLatestEvents` permanently terminates the `Backend.poll` goroutine for the remainder of the `k8s-dqlite` process's lifetime. Because `Backend.poll` is the only writer to every `WatcherGroup`'s update channel, this silently disables:

- all in-flight `etcd` watch streams (apiserver `Watch` RPCs hang forever, returning no events and no errors),
- TTL/lease expiration (kubelet/controller-manager/scheduler leases stop being deleted),
- any future watcher created on the same backend (it is dead-on-arrival).

Compaction, `Range`/`Get`/`Put`, dqlite raft, and gRPC keepalives all keep working, so the process appears healthy from the outside. Symptoms surface as Kubernetes leader-election failures, stale informer caches, and watch RPCs that never return.

We hit this in production on microk8s; impact is "watching is dead until k8s-dqlite is restarted."

## Repro

1. Start `k8s-dqlite` normally (or any HA microk8s cluster).
2. Use `iptables` to block outbound connections toward other nodes (e.g. `sudo iptables -I OUTPUT 1 -p tcp -d NODE_IP --dport 19001 -j DROP` repeat for other node)
3. Observe that the next time anything writes to the database, watchers do not receive the event.
4. From that point on, *no* watch ever delivers an event, even after the database is healthy again.

## Evidence (goroutine pprof)

Comparing two `goroutine?debug=1` snapshots from the same process — one before the issue (`fast`) and one after (`slow`):

| Marker | fast | slow |
|---|---|---|
| Process uptime | 7m | 38m |
| `Backend.poll` (`pkg/backend/internal/backend/watcher.go`) | **1** | **0** |
| `Backend.startWatch.func2` (parent of `poll`) | **1** | **0** |
| `Backend.startWatch.func1` (compaction loop) | 1 | 1 |
| `watchStream.ServeUpdates` parked at `pkg/limited/watch.go:87` | 72 (no age) | 59 (**all** `chan receive, 35 minutes`) |
| `Backend.ttl.func2` parked on `Updates()` (`ttl.go:86`) | `chan receive` (no age) | `chan receive, **35 minutes**` |

Reading: in `slow`, every consumer of `WatcherGroup.Updates()` has been blocked on the same channel for 35 minutes (out of 38 minutes of uptime). The compaction goroutine — the *sibling* of `poll` started in the same `startWatch` — is still alive, so the backend's parent context is not canceled. Yet `poll` is gone. The only way that can happen is the `errors.Is(err, context.DeadlineExceeded)` early-return path in `Backend.poll`.

Newly created watchers in the slow dump have very high goroutine ids but still report `chan receive, 35m` — i.e. they were instantly dead-on-arrival, never received an event, just sat on the channel for as long as the gRPC stream stayed open.

## Root cause

In `pkg/backend/internal/backend/watcher.go`:

```go
func (s *Backend) poll(ctx context.Context) {
    // ...
    for {
        // ...
        events, err := s.getLatestEvents(ctx, s.pollRevision)
        if err != nil {
            if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
                return                       // <-- bug
            }
            logrus.Errorf("fail to get latest events: %v", err)
            continue
        }
        // ...
    }
}

func (s *Backend) getLatestEvents(ctx context.Context, pollRevision int64) ([]*limited.Event, error) {
    watchCtx, cancel := context.WithTimeout(ctx, s.Config.WatchQueryTimeout)
    defer cancel()
    rows, err := s.Driver.After(watchCtx, pollRevision, pollBatchSize)
    // ...
}
```

`getLatestEvents` runs the SQL query under a **per-query** deadline (`WatchQueryTimeout`). When the inner `watchCtx` deadline fires, `Driver.After` returns `context.DeadlineExceeded` and `errors.Is(err, context.DeadlineExceeded)` is `true`, so `poll` returns — even though the *outer* `ctx` from `Backend.Start` (the backend lifetime) is still alive.

Consequences:

- `s.WatcherGroups` is never drained.
- The per-watcher `context.AfterFunc` cleanup runs only when the gRPC stream closes, so existing watchers' channels stay open and unfed.
- `KVServerBridge.Watch` happily continues creating new `WatcherGroup`s; they're added to the same dead set.
- Nothing logs "watcher poll exited"; the failure is silent.

This is the classic "compare against the wrapped sentinel error of a nested timeout" pitfall. The check needs to ask whether the **parent** context is done, not whether *any* context in the chain is done.

## Expected behavior

A transient query-timeout should be retried (with backoff). The poll goroutine must only exit when its parent context is canceled (i.e. on backend shutdown). Watchers should keep getting events as soon as the database recovers.

## Proposed fix (gist)

```go
events, err := s.getLatestEvents(ctx, s.pollRevision)
if err != nil {
    if ctx.Err() != nil {           // backend itself is shutting down
        return
    }
    logrus.WithError(err).Errorf("fail to get latest events")
    select {
    case <-ctx.Done():
        return
    case <-time.After(s.Config.PollInterval):   // small backoff, avoid hot-looping on a sick DB
    }
    continue
}
```

Also worth considering as defense-in-depth:

- A supervisor in `startWatch` that re-runs `poll` while `ctx.Err() == nil`, so even an unforeseen exit (panic, future bug) doesn't take watching out for the rest of the process.
- An `Error`-level log and a counter increment if `poll` returns while `ctx.Err() == nil` (to surface a recurrence loudly).

## Severity

High. Watching is a hard requirement for a working kube-apiserver; the failure is silent and the only mitigation today is to restart `k8s-dqlite` (which only postpones recurrence under load).

strainovic:
## Summary

Fixes #364.

`Backend.poll` was permanently exiting whenever a single `Driver.After` call hit the per-query `WatchQueryTimeout`, which silently disabled all watch event delivery and TTL expiration for the rest of the process's lifetime. This PR makes the poll loop only exit when its **parent** context is canceled (i.e. backend shutdown), and treat per-query timeouts as transient with backoff-and-retry.

## Root cause (short version)

`Backend.getLatestEvents` wraps the SQL call in `context.WithTimeout(ctx, s.Config.WatchQueryTimeout)`. When that inner deadline fires, the returned error matches `errors.Is(err, context.DeadlineExceeded)`. The previous code treated that match as a fatal signal and returned from `poll` — even though the outer backend context was still alive.

Consequences observed in production:

- All `watchStream.ServeUpdates` goroutines parked indefinitely on `for update := range ws.watcherGroup.Updates()` (`pkg/limited/watch.go:87`).
- `Backend.ttl.func2` parked on the same `Updates()` channel — leases stopped expiring.
- Compaction kept running, dqlite stayed healthy, nothing logged the failure → silent breakage.
- A goroutine pprof comparison clearly showed `Backend.poll` present in the healthy snapshot and absent in the broken one, while the sibling compaction goroutine remained in both.

## What changed

`pkg/backend/internal/backend/watcher.go`:

- The poll loop now exits **only** when `ctx.Err() != nil` (backend shutdown). Per-query `context.DeadlineExceeded` and other transient errors are logged and retried.
- A short, cancelable backoff (`PollInterval`) is inserted between failed attempts to avoid hot-looping while the database is sick.
- A loud error log and a `WARNING` are emitted on every failed `getLatestEvents`, so a recurrence is no longer silent.
- (Optional, included if you want belt-and-suspenders) `Backend.startWatch` now supervises `poll`: if `poll` ever returns while the backend context is alive, the supervisor logs at `Error` and restarts it. This converts any future regression of the same class into a self-healing event with telemetry, instead of silent breakage.

`pkg/backend/internal/backend/backend.go`:

- Tiny refactor of `startWatch.func2` to host the supervisor described above.

## Why `ctx.Err()` instead of `errors.Is(err, context.DeadlineExceeded)`

`context.DeadlineExceeded` is a **value**, not a context. Two different `context.WithTimeout`s that fire produce errors that both satisfy `errors.Is(err, context.DeadlineExceeded)`, but they mean very different things to the caller. The only reliable way to distinguish "my timeout fired" from "my parent's timeout/cancel fired" is to inspect the parent's `ctx.Err()` directly. This is a well-known footgun (referenced from Go's own documentation) and the lesson generalizes — the codebase should prefer `ctx.Err()` checks over `errors.Is(..., context.DeadlineExceeded)` whenever inner timeouts are involved.

## Compatibility / risk

- No protocol, schema, or on-disk format changes.
- No public API changes.
- Behavior change is strictly "watch keeps working in cases where it previously broke". A misconfigured `WatchQueryTimeout` that's too short will now manifest as logs (`fail to get latest events`) and a small CPU/IO uptick from retries, instead of the previous silent freeze. This is the intended trade-off and is observable.
- HA upgrade is safe; rolling restart node-by-node is sufficient.


dwbfox:
bumping this thread to see if there's any cycles for additional eyes on this issue. It seems a microk8s node becomes sluggish or unresponsive once there's a network disruption to the dqlite port (19001). Simply blocking incoming connections to the port so there is a temporary drop in connection to a node will be able to trigger the condition once the connectivity is restored.

louiseschmidtgen (commented):
Thank you @strainovic, 
can you please rebase this PR on main as I've fixed the integration test suite in my last commit.
The patch looks good from my end.

strainovic:
Hi @louiseschmidtgen,

Done.

Thank you!

louiseschmidtgen (approved):
lgtm

